PR #2, opened 2026-06-24
fix(sessions): stop dropping sessions when the server returns short pages

AIalliAI:
## What

`loadAllHermesSessions` ended its pagination loop as soon as a page returned fewer than `limit` rows (`rows.length < limit`). With `include_children=true` the gateway can expand rows server-side and return a short page while `has_more`/`total` still indicate more sessions remain — so the session picker silently loaded only a prefix of the user's sessions.

## Changes

Replace the `rows.length < limit` early-stop with an exported `shouldStopSessionPaging` predicate that honors the explicit `has_more`/`total` signals and only stops on an empty page or when no more are reported. The existing 10-page cap and empty-page guard still bound the loop.

## Verification

- `npm test` is green. Adds a unit test for the predicate across the meaningful branches (short page with more remaining, empty page, completed, no signals).
- Negative control: the old inline condition stops on a 300-row page (limit 500) while `total: 1200, has_more: true`, dropping 400 sessions; the new predicate keeps paging. Reverting the source fails the new test.

## Notes

- Bounded: the loop still hard-caps at 10 pages × 500, and an empty page always stops even under `has_more: true`, so there is no infinite-loop risk.
- When `has_more` and `total` disagree, the predicate trusts `has_more` (keeps paging), costing at most one extra fetch that returns empty and then stops.

### DIFF
--- a/extension/lib/common.mjs
+++ b/extension/lib/common.mjs
@@ -798,3 +798,15 @@ export function extractAssistantText(payload) {
 export function encodeSessionId(sessionId = DEFAULT_SETTINGS.sessionId) {
   return encodeURIComponent(String(sessionId || DEFAULT_SETTINGS.sessionId).trim() || DEFAULT_SETTINGS.sessionId);
 }
+
+// Decide whether session pagination should stop after consuming a page.
+// `offset` is the running total of rows already consumed (including this page).
+// A page shorter than `limit` is NOT a stop signal on its own: servers can
+// return short pages (e.g. when include_children expands rows) while more
+// pages remain, so we honor the explicit has_more / total signals instead.
+export function shouldStopSessionPaging({ rowCount = 0, offset = 0, total = 0, hasMore = false } = {}) {
+  if (!rowCount) return true;
+  if (hasMore) return false;
+  if (total && offset < total) return false;
+  return true;
+}
--- a/extension/sidepanel.js
+++ b/extension/sidepanel.js
@@ -22,6 +22,7 @@ import {
   reasoningEffortShortLabel,
   renderMarkdown,
   safeTab,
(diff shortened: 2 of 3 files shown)


AIalliAI:
Already on `main` as of v0.1.2 (`4a79950`): the exported `shouldStopSessionPaging` predicate and its wiring into `loadAllHermesSessions` are present and covered by the 'session paging helper continues on explicit has_more or total boundary' test. Closing as superseded.

abundantbeing:
## Hermes Agent Review

## Summary

The change is narrowly scoped and the new predicate fixes the reported regression: short non-empty pages no longer terminate pagination when `has_more` or `total` indicates more sessions remain. I found no blocking correctness or security issue in the supplied diff.

Not approving based on this event: the PR body’s “npm test is green” claim is untrusted, and no CI/check output is included in the event data.

## Blocking issues

None found in the supplied diff.

## Suggestions

- Add a small integration-style test around `loadAllHermesSessions`, not only `shouldStopSessionPaging`, to verify the loop actually performs the next fetch after a short page and advances `offset` as expected.
- Consider hardening pagination metadata parsing later if gateway responses may use string booleans, e.g. `"false"` for `has_more`; `Boolean("false")` evaluates to `true`. This is pre-existing behavior, not introduced here.

## Checks to run

- `npm test`
- Focused test for `tests/common.test.mjs`
- Browser/manual smoke against a gateway with more than one page of sessions and `include_children=true`, confirming the session picker shows sessions beyond the first short page.
- If available: extension lint/build/package validation before release.

---

Reviewed signature: `f52caeb1ca0929966b4fa9b0ec20af576bfc07e5863798f3b9c1a30f844113d4`

PR #3, opened 2026-06-24
fix(stream): stop double-counting OpenAI-style streams that end with a full message

AIalliAI:
## What

`appendOpenAiChunkText` fell back to `delta.content || message.content` and always **appended** the result. When a gateway streams incremental deltas and then a terminal chunk carrying the full assistant message (some OpenAI-compatible servers do), the full message was appended on top of the accumulated deltas, duplicating the reply end to end (`"Hello world"` → `"Hello worldHello world"`).

## Changes

Distinguish the two cases: append incremental `delta.content`, but **replace** the accumulator when a chunk instead carries the full `message.content`. Move the helper into `lib/common.mjs` (exported) so it can be unit-tested; the sole call site in `sidepanel.js` imports it.

## Verification

- `npm test`, `npm run check:js`, and `npm run check:manifest` are green. Adds a regression test for the delta-then-full-message sequence plus `[DONE]`/empty-chunk handling.
- Negative control: the old implementation yields `"Hello worldHello world"` and fails the new test; the new implementation yields `"Hello world"`.

## Notes

- "Replace on full message" treats a terminal full-message chunk as the complete reply (a superset of the deltas), which is what such a chunk means by construction. It is strictly better than the previous behavior for both the delta-then-full and the cumulative-`message.content` server shapes.

### DIFF
--- a/extension/lib/common.mjs
+++ b/extension/lib/common.mjs
@@ -795,6 +795,21 @@ export function extractAssistantText(payload) {
   return '';
 }
 
+// Accumulate text from an OpenAI-style streaming SSE chunk.
+// Incremental `delta.content` is appended; a chunk that instead carries the
+// full `message.content` REPLACES the accumulated text. Appending the full
+// message (as the previous implementation did) double-counts the answer when a
+// stream emits deltas and then a terminal full-message chunk.
+export function appendOpenAiChunkText(event = {}, finalText = '') {
+  if (event?.data === '[DONE]') return finalText;
+  const choice = (event?.json || {}).choices?.[0] || {};
+  const delta = choice.delta?.content;
+  if (delta) return `${finalText}${delta}`;
+  const message = choice.message?.content;
+  if (message) return String(message);
+  return finalText;
+}
+
 export function encodeSessionId(sessionId = DEFAULT_SETTINGS.sessionId) {
   return encodeURIComponent(String(sessionId || DEFAULT_SETTINGS.sessionId).trim() || DEFAULT_SETTINGS.sessionId);
 }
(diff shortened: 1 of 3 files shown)


AIalliAI:
Already on `main` as of v0.1.2 (`4a79950`): `appendOpenAiChunkText` (now in `lib/common.mjs`) with append-delta / replace-on-full-message is present and covered by the 'OpenAI stream chunks append deltas and preserve final message payloads' test. Closing as superseded.

abundantbeing:
## Hermes Agent Review

## Summary

Low-risk, focused fix. The stream accumulator change correctly separates incremental `delta.content` from terminal/full `message.content`, and the call site import looks clean. No code-level blocking issue found in the diff.

I’m not approving because the visible GitHub check state is not green: the only check run I could see is `Kilo Code Review` with `action_required` due insufficient credits, and no CI status proving `npm test` / JS checks passed.

## Blocking issues

None found in the changed code.

Release gate: don’t merge until required GitHub checks are green or a maintainer explicitly accepts the missing CI evidence.

## Suggestions

- `extension/lib/common.mjs:806-809`: consider checking string presence by type, not truthiness:
  - `typeof delta === 'string'`
  - `typeof message === 'string'`
  This preserves intentional empty terminal messages and avoids accidentally stringifying non-string content into `[object Object]`.

- `tests/common.test.mjs`: add a direct cumulative `message.content` regression case, e.g. `message("Hel")` then `message("Hello")` should end as `"Hello"`. The current test covers delta-then-full-message, which is the main bug.

## Checks to run

- `npm run verify`
- Confirm GitHub branch protection / required checks are green before merge.
- Optional manual smoke: stream against an OpenAI-compatible gateway that emits deltas plus a terminal full message and verify the side panel shows the reply once.

---

Reviewed signature: `5698bb55527d92cd6c135b30911e686edea84db237af807e0b98cc6040370f64`

PR #4, opened 2026-06-24
fix(markdown): render GFM tables with single-dash dividers

AIalliAI:
## What

`isTableDivider` required at least three dashes per column (`-{3,}`), so a valid GFM table whose divider row is written compactly (e.g. `|-|-|`) was not recognized and fell through to the paragraph renderer — surfacing the raw pipes in the assistant reply.

## Changes

Loosen the divider pattern to one-or-more dashes (`:?-+:?`) while keeping optional alignment colons. Horizontal rules and multi-dash dividers are unaffected (HR is matched earlier and has no pipes).

## Verification

- `npm test` is green. Adds a regression test that the single-dash divider renders a real `<table>` and the raw `|-|-|` does not leak as a paragraph.
- Negative control: reverting the source fails the new test.

## Notes

- Matches GFM, which accepts one-or-more dashes per column. Multi-column tables are still required for promotion (the divider must have ≥2 columns), so a lone `|-|` is not treated as a table.

### DIFF
--- a/extension/lib/common.mjs
+++ b/extension/lib/common.mjs
@@ -311,7 +311,7 @@ function isHorizontalRule(line = '') {
 }
 
 function isTableDivider(line = '') {
-  return /^\s*\|?\s*:?-{3,}:?\s*(\|\s*:?-{3,}:?\s*)+\|?\s*$/.test(line);
+  return /^\s*\|?\s*:?-+:?\s*(\|\s*:?-+:?\s*)+\|?\s*$/.test(line);
 }
 
 function tableCells(line = '') {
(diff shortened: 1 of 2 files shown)


AIalliAI:
Already on `main` as of v0.1.2 (`4a79950`): the loosened `isTableDivider` (`:?-+:?`) is present and covered by the table test. Closing as superseded.

abundantbeing:
## Hermes Agent Review

## Summary
Small, focused markdown parser fix. The regex change correctly allows one-or-more dashes in GFM table divider cells, and the added regression test covers the reported compact divider case.

## Blocking issues
None found in the provided diff.

Not an approval: the PR body claims `npm test` is green, but no trusted CI/check output was included in the event data.

## Suggestions
- Add one more negative regression if possible: ensure a single-column compact divider like `|-|` still does not promote to a table, since the PR notes that behavior as intentional.
- Consider adding alignment coverage for compact dividers, e.g. `|:-|-:|`, to protect the optional colon behavior.

## Checks to run
- `npm test`
- If available for this repo: lint/build checks for the browser extension bundle.

---

Reviewed signature: `116b58b44915115d35285a5a1e9ea98363e40131f326b8784891d7676888df94`

PR #15, opened 2026-06-25
Add a remote-dashboard gateway mode (WebSocket, no API server)

ptaranat:
Adds a **Dashboard (WebSocket)** option for reaching a remote Hermes when the only thing exposed is the OAuth-gated dashboard, with no API server. It talks to the dashboard's `/api/ws` socket and authenticates with a single-use ws-ticket minted from a signed-in dashboard tab, since the extension can't send the dashboard cookie cross-origin. No API key, no API server.

The gateway picker stays just Local/Remote. In Remote the transport is inferred from the API key field: a key uses your `remote-api` path, blank uses the dashboard WebSocket. Same `gatewayMode` underneath. I also fixed the setup-help overflow and trimmed the copy.

Tests included. Image upload and the skills/profiles lists don't work over the dashboard socket (REST-only, CORS-blocked). Totally fine if it's more than you want to carry upstream.

## Screenshot

chrome_9jfeRN9HFl

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ In the side panel first-run screen:
 1. Click **Connect to Hermes** and approve locally if your Hermes Desktop/gateway supports the approval flow.
 2. If approval is not available yet, click **Manual setup**.
 3. Enter:
-   - Gateway mode: **Local API server** or **Remote API server**.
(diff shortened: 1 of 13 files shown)


ptaranat:
@abundantbeing whoops I didn't realize you merged 6e475b5
I'll rework this to fit your model.

ptaranat:
Reworked.

PR #18, opened 2026-06-27
style: add ESLint and Prettier configuration

iruzen-dono:
## Summary

Add ESLint (flat config, v9) and Prettier configuration to the project, plus matching npm scripts. This gives contributors and editors standard tooling for catching issues early and keeping a consistent code style.

## Changes

**`eslint.config.js`** — ESLint flat config with:
- `eslint:recommended` as the base rule set
- Browser (`chrome.*`, `document`, `fetch`, etc.) and Node.js (`process`, `Buffer`, etc.) globals scoped per file pattern
- Sensible extras: `no-var` (error), `prefer-const` (warn), `prefer-template` (warn), `no-throw-literal` (warn)
- Unused-vars warns with `_` prefix exemption
- Ignores `dist/`, `artifacts/`, `node_modules/`

**`.prettierrc`** — matching the project's existing style (single quotes, semicolons, 2-space indent, trailing commas, 120 print width)

**`package.json`** — four new scripts:
- `npm run lint` / `npm run lint:fix` (ESLint)
- `npm run format:check` / `npm run format` (Prettier)

**`CONTRIBUTING.md`** — updated with lint/format instructions

## Notes

- ESLint v9 uses flat config natively, no `.eslintrc.*` legacy format needed
- Config is intentionally permissive (warnings, not errors, for most styles) — catches real issues without blocking development
- After `npm install`, the tools are ready to use

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,9 @@
+{
+  "semi": true,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "all",
+  "printWidth": 120,
+  "arrowParens": "always",
+  "endOfLine": "lf"
+}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,8 @@ Thanks for helping improve Hermes Browser Extension.
 
 - Use Node.js 20+.
 - Run `npm run verify` before opening a PR.
+- Run `npm run lint` and `npm run format:check` to catch style and syntax issues
+  early. Use `npm run format` to auto-fix formatting.
 - Keep v0.1 read-only unless a proposal explicitly changes the permission model.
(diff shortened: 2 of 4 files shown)


abundantbeing:
Thanks — merged the static-checking contribution and then adjusted the final landing so it is safe for the repo baseline.

What changed after merge:
- ESLint now runs with zero errors on the current extension/runtime globals.
- `package-lock.json` is committed, so `npm ci` works for fresh installs.
- Prettier enforcement was intentionally left out of this landing because the current repo would need a broad formatting pass; I do not want hidden repo-wide formatting churn inside a tooling PR.

Verified on `main` with:
- `npm ci`
- `npm run lint` (0 errors; existing unused-helper warnings only)
- `npm run verify`
- `npm run build`

PR #19, opened 2026-06-27
fix: handle unloaded and failed context states in DOM chip

iruzen-dono:
## Summary

Fixes the "DOM chip says 0 chars even on a normal page" issue by distinguishing three states the chip can be in: **loading**, **restricted/error**, and **normal operation**.

## The problem

Previously the chip displayed `📎 DOM · 0 chars · ~0 tok` in all states where text was 0 — even when the page context hadn't loaded yet, when the page was restricted, or when the context capture failed. This conflated three distinct situations:

1. **Initial load** — panel opened but `refreshContext()` hasn't completed yet
2. **Restricted page** — `chrome://`, `file://`, `about:blank`, or other non-injectable URLs
3. **Capture failure** — stale content script, tab not ready, or connection error
4. **Legitimate empty page** — page genuinely has no readable text

## The fix

**`extension/sidepanel.html`** — initial placeholder changed from `📎 DOM · 0 chars · ~0 tok` to `📎 Loading...`

**`extension/sidepanel.js`** (`renderContextWindow`) — chip now reflects the real state:

| Chip display | When |
|---|---|
| `📎 Loading...` | `pageContext` is null/undefined (not yet captured) |
| `📎 Restricted · N/A` | `pageContext.restricted === true` |
| `📎 Error · N/A` | `pageContext.ok === false` |
| `📎 DOM · X chars · ~Y tok` | Normal — page context captured successfully |

Also improves the context preview tooltip with better title attributes for each state.

## Testing

- **Normal page** (e.g. github.com) — shows `📎 DOM · X chars · ~Y tok` as before
- **`about:blank`** or new tab — shows `📎 Restricted · N/A`
- **`chrome://extensions/`** — shows `📎 Restricted · N/A`
- **Before first context refresh** — briefly shows `📎 Loading...`

### DIFF
--- a/extension/sidepanel.html
+++ b/extension/sidepanel.html
@@ -91,7 +91,7 @@
         <form id="composer" class="composer">
           <label class="composer-label" for="promptInput">ASK HERMES</label>
           <button id="contextChip" class="context-chip" type="button" aria-expanded="false">
-            <span id="contextChipLabel">📎 DOM · 0 chars · ~0 tok</span>
+            <span id="contextChipLabel">📎 Loading...</span>
             <strong>preview</strong>
           </button>
           <div id="contextPreview" class="context-preview" hidden></div>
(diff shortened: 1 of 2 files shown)


abundantbeing:
Thanks — merged this. I also added a follow-up `contextChipSummary` helper + regression test so the loading, restricted, error, and captured chip states are covered by the test suite now.

Verified on `main` with:
- `npm ci`
- `npm run lint` (0 errors; existing unused-helper warnings only)
- `npm run verify`
- `npm run build`

PR #21, opened 2026-06-27
feat(sidepanel): multi-tab context picker with user-selectable tabs

iruzen-dono:
## Feature : multi-tab context picker

Adds a tab picker dropdown alongside the context chip, letting users select specific browser tabs to include in the Hermes prompt context — no longer forced to include all open tabs.

### What changed

- **extension/sidepanel.html** — 📑 tab picker button with badge next to context chip
- **extension/sidepanel.css** — full tab-picker component styles (dropdown, items, checkboxes, favicons, search, action buttons)
- **extension/lib/common.mjs** — `buildHermesPrompt` accepts optional `selectedTabs` param; when fewer tabs are selected than available, lists them in UNTRUSTED_BROWSER_CONTEXT
- **extension/sidepanel.js** — `selectedTabs` state, `renderTabPicker()`, `updateTabPickerButton()\', toggle/outside-click handlers, wired into `refreshContext` and `askHermes`

### UX

- Badge shows selected count; button highlights when a non-default selection is active
- Searchable dropdown with Select All / Deselect All
- Closing picker does not reset selection — choice persists until changed or session ends
- null = all tabs (no-op), array = user-filtered subset

### Testing

- `npm run check:js` ✓ (14 files pass)
- `npm run test` ✓ (78/78 pass)
- The lint failure is pre-existing (`@eslint/js` missing from main too)

### DIFF
--- a/extension/lib/common.mjs
+++ b/extension/lib/common.mjs
@@ -1003,7 +1003,7 @@ function formatMeta(meta = {}) {
   return parts.join('\n\n');
 }
 
-export function buildHermesPrompt({ userText, activeTab, tabs, pageContext, settings = DEFAULT_SETTINGS }) {
+export function buildHermesPrompt({ userText, activeTab, tabs, pageContext, selectedTabs, settings = DEFAULT_SETTINGS }) {
   const mergedSettings = { ...DEFAULT_SETTINGS, ...settings };
   const limit = contextCharLimit(mergedSettings.contextDepth);
   const selectedText = mergedSettings.includeSelectedText ? redactSensitiveText(pageContext?.selectedText || '') : '';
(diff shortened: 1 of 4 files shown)


abundantbeing:
## Hermes Agent Review

Summary
Multi-tab picker UI added next to the context chip, with search/select-all/deselect-all and badge state. The picker stores selectedTabs and passes that to buildHermesPrompt so the prompt can note when a subset is in use. Tests noted in PR description: npm run check:js and npm run test (78/78 passing).

Blocking issues
Selected tabs never actually limit the tab list that reaches Hermes. buildHermesPrompt still renders tabsText from the full tabs array and the only thing selectedTabs does is append a sentence saying the user selected some tabs. Because askHermes still passes context.tabs (all tabs) and no filtering happens before formatting, the feature does nothing except change the badge state, so users cannot exclude unselected tabs from the prompt despite the UI claiming otherwise. Please make the chooser filter the tab list that gets serialized (e.g. derive tabsText from the selected subset or pass a filtered tabs array to buildHermesPrompt and/or refreshContext) so the selection drives the actual context.

Suggestions
Clamping selectedTabs against the latest currentContext.tabs on refresh would keep the badge/count in sync when tabs close or the window changes; drop entries for missing tabIds and reset to null if nothing remains so the picker doesn’t show stale counts.

Checks to run
npm run check:js (pass noted in PR body)
npm run test (78/78 pass noted in PR body)

---

Reviewed signature: `abbb038f69e80403bcb293f4d731d6cfe7584d3f4eadf04ba0e2a842371d99d2`

iruzen-dono:
Blocking issue fixed in 907a589. Two changes:

1. **buildHermesPrompt now filters tabs by selectedTabs.** When `selectedTabs` is an array with content, `summarizeTabs()` receives the filtered subset instead of the full `tabs` array. The prompt context now actually reflects the user's tab selection instead of just noting the count in a sentence.

2. **Tab picker selection ID mismatch fixed.** `safeTab()` returns `{ id, active, pinned, ... }` (Chrome's `tab.id`), but the picker code used `tab.tabId` everywhere — resulting in `undefined === undefined` comparisons that always evaluated `true`, making toggling ineffective. All `tab.tabId` references changed to `tab.id`.

PR #22, opened 2026-06-28
fix: always run full Hermes model discovery

abundantbeing:
Summary:
- keep the connection auth probe but stop passing the /v1/models alias payload into model sync
- force the normal model discovery chain on connect so /api/model/options and session fallback can populate real providers/models
- make manual model refresh request a fresh registry fetch

Verification:
- npm test
- npm run check:js
- npm run check:manifest
- npm run build
- npm run verify

### DIFF
--- a/extension/lib/model-discovery.mjs
+++ b/extension/lib/model-discovery.mjs
@@ -91,6 +91,8 @@ export async function discoverModelsFromRegistry({
         seen.add(dedupeKey);
         const modelCaps = caps[modelId] || {};
         const uiId = slug ? `${slug}::${modelId}` : modelId;
+        const entryContext = Number(entry?.context_length || entry?.contextTokens || 0) || 0;
+        const capsContext = Number(modelCaps?.context_length || 0) || 0;
         models.push({
           id: uiId,
           rawModelId: modelId,
@@ -98,7 +100,7 @@ export async function discoverModelsFromRegistry({
           provider: slug || deriveProviderFromModelId(modelId),
           providerLabel,
           description: provider?.warning || provider?.source || '',
-          contextTokens: Number(entry?.context_length || entry?.contextTokens || 0) || 0,
+          contextTokens: entryContext || capsContext || 0,
           fast: typeof modelCaps.fast === 'boolean' ? modelCaps.fast : undefined,
           reasoning: typeof modelCaps.reasoning === 'boolean' ? modelCaps.reasoning : undefined,
           authenticated: provider?.authenticated !== false,
--- a/extension/sidepanel.js
+++ b/extension/sidepanel.js
@@ -2081,7 +2081,7 @@ function applySelectedModel(selectedId, { persist = true, keepOpen = false } = {
   }
 }
 
-async function loadModels({ quiet = false, payload = null } = {}) {
+async function loadModels({ quiet = false, payload = null, refresh = false } = {}) {
   try {
     let data = payload;
     let registryModels = [];
@@ -2102,7 +2102,7 @@ async function loadModels({ quiet = false, payload = null } = {}) {
     if (data) {
       registryModels = normalizeHermesModels(data, settings.model);
     } else {
-      const registryResult = await discoverModelsFromRegistry({ apiFetch, readJsonResponse });
+      const registryResult = await discoverModelsFromRegistry({ apiFetch, readJsonResponse, refresh });
       if (registryResult.ok && registryResult.models.length) {
         registryModels = normalizeHermesModels(registryResult.models, settings.model);
         registrySource = 'registry';
@@ -4060,7 +4060,7 @@ async function testConnection() {
     const modelsResponse = await apiFetch('/v1/models', { method: 'GET' });
     const modelsPayload = await readJsonResponse(modelsResponse);
     if (!modelsResponse.ok) throw new Error(`Health OK, auth/model probe failed (${modelsResponse.status}): ${JSON.stringify(modelsPayload).slice(0, 500)}`);
-    await loadModels({ quiet: true, payload: modelsPayload });
+    await loadModels({ quiet: true });
     await loadSkills({ quiet: true });
     await loadProfiles({ quiet: true });
 
@@ -4182,7 +4182,7 @@ function bindEvents() {
   els.stopButton?.addEventListener('click', stopCurrentTurn);
   els.voiceButton?.addEventListener('click', toggleVoiceDictation);
   els.checkUpdatesButton?.addEventListener('click', checkForUpdates);
-  els.refreshModelsButton.addEventListener('click', () => loadModels());
+  els.refreshModelsButton.addEventListener('click', () => loadModels({ refresh: true }));
   els.refreshProfilesButton?.addEventListener('click', () => loadProfiles());
   els.profileSelect?.addEventListener('change', () => applySelectedProfile(els.profileSelect.value));
   els.refreshAgentsButton?.addEventListener('click', () => loadAgents());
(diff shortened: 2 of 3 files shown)


abundantbeing:
## Hermes Agent Review

# GitHub Review: PR #22 — fix: always run full Hermes model discovery

**Repo:** abundantbeing/hermes-browser-extension
**Author:** abundantbeing
**Branch:** (not shown in event data)

## Summary

This PR decouples auth validation from model discovery in `extension/sidepanel.js`. The `/v1/models` endpoint is kept only as an auth probe; its response payload is no longer fed into `loadModels`. Instead, `testConnection` now calls `loadModels({ quiet: true })` with no payload, forcing the normal discovery chain (`/api/model/options` + session fallback) to run on connect. A new `refresh` flag is plumbed through to `discoverModelsFromRegistry` so the manual refresh button forces a fresh registry fetch. A new structural test in `tests/common.test.mjs` asserts the wiring.

The intent is clear and the change is narrow and well-scoped. No secrets, no eval, no new network surface beyond what already existed.

## Blocking issues

None. The diff is small, surgical, and preserves the auth check (the `/v1/models` request still throws on non-OK).

## Suggestions

1. **Test is a code-shape test, not a behavior test.** The new test in `tests/common.test.mjs` only regex-matches `sidepanel.js` source. It will pass even if `loadModels` is refactored into a state where it silently no-ops on connect. Consider an integration-style test that mocks `apiFetch` and asserts that after `testConnection()`, `discoverModelsFromRegistry` (or its successor) is invoked with `{ refresh: false }` and that the `/v1/models` response is *not* passed as `payload`.

2. **`refresh` flag only reaches the registry path.** When `loadModels` is called with a `payload`, `refresh` is silently ignored. That is fine semantically (payload = "use this data, no fetch"), but worth a short comment in `loadModels` so future readers don't add a duplicate fetch.

3. **Auth probe endpoint choice.** Using `/v1/models` purely as an auth probe is unusual; if that endpoint is ever deprecated or returns 200 with an empty list for unauthenticated requests, the probe will mis-report success. Pre-existing, not introduced here, but flagging since the PR is reorganizing this exact code path. A dedicated `/health` or `/me` endpoint would be safer long-term.

4. **Silent fallback risk.** If `/api/model/options` returns ok-but-empty and the session fallback also yields nothing, the user now sees no models after a successful auth probe (where previously the `/v1/models` payload would at least populate *something*). Verify the empty-state UI handles this gracefully and consider logging a warning when both discovery paths return zero models post-auth.

5. **`payload` parameter is now dead in the connect path.** Worth grepping for other call sites of `loadModels({ ..., payload: ... })` to confirm none still rely on the old behavior. The new test only asserts the connect path; it doesn't check the rest of the codebase.

6. **One additional assertion to harden the test:** `assert.match(source, /function loadModels\(\{ quiet = false, payload = null, refresh = false \}\)/)` would catch a regression where someone removes the `refresh` parameter.

## Checks to run

The PR body claims the following passed locally:

- `npm test`
- `npm run check:js`
- `npm run check:manifest`
- `npm run build`
- `npm run verify`

**I cannot confirm these from the event data.** No CI run URL, no log excerpt, no status check payload is included. Before merging, verify the checks are actually green on the PR's commit(s) (not just claimed in the description) and that the new test is included in the `npm test` run.

Also worth running manually:

- `git grep -n "loadModels(" extension/` — confirm no other call sites pass `payload:` that would silently change behavior
- `git grep -n "discoverModelsFromRegistry"` — confirm the `refresh` parameter is handled at every call site, not just the one updated
- A real Connect flow against a live Hermes backend: confirm models populate and the manual refresh button produces a fresh list

## Verdict

Not blocking. Small, sensible fix with a clear motivation. Ship it after the CI checks above are confirmed green and the dead-`payload` grep comes back clean.

---

Reviewed signature: `0d97f7769dd18f4fb59c9679a5982e679c187751212270d61ab04291175a8cac`

abundantbeing:
Verification follow-up from the automated review:

Commands re-run on the PR head:

- `git grep -n "loadModels(" extension/`
- `git grep -n "discoverModelsFromRegistry"`
- `npm test` → `94 passed`
- `npm run check:js` → passed
- `npm run check:manifest` → `Manifest OK: Hermes Browser Extension 0.1.6`
- `npm run build` → built `dist/`
- `npm run verify` → passed

Relevant grep result: the connect/test path now calls `loadModels({ quiet: true })` without the `/v1/models` payload, and the only registry discovery call is `discoverModelsFromRegistry({ apiFetch, readJsonResponse, refresh })`. Manual model refresh calls `loadModels({ refresh: true })`.

The remaining `payload:` call is the intentional remote WebSocket model-list path, not the local API-server connect path.

abundantbeing:
## Hermes Agent Review

```markdown
## Summary
PR fixes a real bug: `testConnection()` was passing the `/v1/models` alias payload directly into `loadModels`, short-circuiting the registry → `/api/model/options` → session-fallback discovery chain. The change keeps the auth probe but discards the alias payload so the full discovery chain runs. Also adds a `context_length` fallback from `capabilities` for bare-string model entries, and threads a `refresh` flag from the manual refresh button.

## Blocking issues
1. **`refresh` is plumbed but not shown consumed.** `sidepanel.js` now calls `discoverModelsFromRegistry({ apiFetch, readJsonResponse, refresh })`, but the `model-discovery.mjs` diff only touches the `contextTokens` line — it does not show `refresh` being added to the function signature or destructured. Either the diff is incomplete (and `refresh` actually drives a cache-bust / `cache: 'no-store'` somewhere not shown) or the param is silently dead and the manual refresh button only works because each click is already a fresh fetch. This needs to be confirmed against the full file before merge — if the param is unused, either implement it or drop it from the call site to avoid misleading readers.

2. **No CI evidence shown.** PR body lists `npm test`, `npm run check:js`, `npm run check:manifest`, `npm run build`, `npm run verify` under Verification, but the event payload includes no check run results. Per the review rules I will not treat those as passed until the checks actually appear in the event data.

## Suggestions
- The new connection-test assertion is a regex over the source string rather than a behavioral test. It will catch a revert of the sidepanel wiring but won't catch a regression where `discoverModelsFromRegistry` is changed in a way that re-bypasses the chain. If feasible, add a small behavioral test that stubs `apiFetch` and asserts `loadModels({ quiet: true })` (no payload) triggers the registry call path.
- `entryContext || capsContext || 0` silently collapses "unknown" and "zero" into the same path. Context length of 0 is implausible in practice, so the fallback is fine, but consider a debug `console.debug` when no context is resolvable — it'll help the next person chasing a "model shows 0 context" report.
- `testConnection()` now does the auth probe plus the full discovery chain, which means more network round-trips on connect. Worth a line in the PR description so users on slow links aren't surprised; consider short-circuiting skills/profiles loads if the registry fetch fails.
- The `refresh` button still fires `loadModels({ refresh: true })` even when no registry result will be returned (e.g., discovery returns empty). If `refresh` ends up meaning "bypass any internal cache", make sure the no-models path surfaces a quiet toast or status — right now `quiet: false` is the default for the refresh button, which is correct, just confirming.

## Checks to run
- Open `extension/lib/model-discovery.mjs` and confirm `refresh` is destructured and actually changes behavior (cache bust, `cache: 'no-store'`, or equivalent). If it does not, fix the wiring before merge.
- `npm test` — confirm the two new tests pass and no other test regresses (especially the existing `discoverModelsFromRegistry flattens /api/model/options provider inventory` test).
- `npm run verify` end-to-end — this is the bundled gate the PR claims, treat its result as authoritative over the individual commands.
- Manual smoke: load unpacked extension → connect to a Hermes backend that has providers the alias `/v1/models` does not surface → confirm those providers now appear → click the refresh button and confirm a new request fires.
- Confirm `/v1/models` is still being fetched in `testConnection` (so auth validation still happens before the full discovery chain runs) and that a failing `/v1/models` still aborts the connect flow.
```

Nothing else blocks merge once the `refresh` wiring in `discoverModelsFromRegistry` is confirmed and the `npm run verify` check is actually green in the event data.

---

Reviewed signature: `3c6852e6166342fb14615adfde928722f12678a54a71b3350ce7cffd964f0a05`

abundantbeing:
Folded the model-discovery fix into main in 660f07d with an additional API-server fallback for one-row /v1/models responses. Closing this branch because it is now behind main and would drop newer composer/update work if merged directly.

PR #25, opened 2026-07-04
feat: add custom model sources for user-configured OpenAI-compatible endpoints

rahlquist:
## Summary

- Adds support for user-configured external OpenAI-compatible API endpoints as model sources
- Enables model discovery from LAN-only inference servers (llama-swap, ollama, vLLM, etc.)
- Switches model discovery from a linear fallback chain to an aggressive multi-source merge

## Motivation

Users running local inference backends (llama-swap, ollama, vLLM) behind OpenAI-compatible endpoints had no way to surface those models in the extension's model picker. The gateway's `/v1/models` only reflects the active provider, and the structured `/api/model/options` registry may not include custom providers. This PR gives users a settings textarea to add any number of external endpoints, which are probed independently and merged into the picker alongside gateway-provided models.

### `extension/lib/model-discovery.mjs`
- **`discoverModelsFromExternalSources()`** — Probes user-configured OpenAI-compatible `/v1/models` endpoints. Supports string-array, `{data:[...]}`, and `{models:[...]}` response formats. Model entries get a hostname-derived provider slug prefixed to their ID (e.g. `wimpy::gemma-3-12b`).
- **`mergeModelsByRawId()`** — Merges model arrays from multiple discovery sources, deduplicating by `rawModelId`. Earlier sources (registry > dashboard > v1 > external) win on conflicts.
- **`mergeModelsWithRegistry()`** — Fixed to fall back from `rawModelId` to `id` when `rawModelId` is absent, preventing silent drops.

### `extension/sidepanel.js`
- **`loadModels()`** — Rewritten from an exclusive fallback chain to an aggressive multi-source gather. Each source is probed independently (registry, dashboard, /v1/models, custom endpoints) and results are merged by `rawModelId`.
- **Settings sync** — `syncSettingsForm()` and `saveSettingsFromForm()` now read/write `customModelSources` from the new textarea.
- **Elements** — Added `customModelSourcesInput` element reference.

### `extension/sidepanel.html`
- Added "Custom model sources" settings section with a `<textarea>` accepting one URL per line.

### `extension/sidepanel.css`
- Monospace 12px font for the custom model sources textarea.

### `extension/lib/common.mjs`
- Added `customModelSources: []` to `DEFAULT_SETTINGS`.

## Test Plan

- [x] All 139 existing unit tests pass
- [x] `npm run build` produces clean dist/
- [x] `discoverModelsFromExternalSources(['http://wimpy:8080/v1'])` returns 21 models (verified via CDP in-test and live)
- [x] `mergeModelsByRawId([gateway_batch, external_batch])` correctly deduplicates and merges (23 unique from 2+21)
- [x] End-to-end browser test: extension connects to gateway, model picker shows both `hermes` (gateway) and `wimpy` (custom source) provider tabs
- [x] Message send/receive works through the gateway

## Setup Requirements

For this feature to work, the Hermes gateway must be started with:
- `API_SERVER_ENABLED=true`
- `API_SERVER_CORS_ORIGINS=chrome-extension://<extension-id>`
- CORS origins must include the extension's `chrome-extension://` origin (the gateway blocks cross-origin requests by default)

## Notes

The `discoverModelsFromExternalSources` function uses `globalThis.fetch` directly (not `apiFetch`), so it does not require gateway authentication — it probes the configured endpoints independently. This means external sources are discovered even when the gateway is unreachable, as long as the configured endpoints are reachable from the browser context.

### DIFF
--- a/extension/lib/common.mjs
+++ b/extension/lib/common.mjs
@@ -57,6 +57,7 @@ export const DEFAULT_SETTINGS = Object.freeze({
   panelResidencyMode: 'tab-attached',
   maxTabs: 12,
   maxLocalMessages: 40,
+  customModelSources: [],
 });
 
 export const HERMES_BROWSER_SYSTEM_PROMPT = `You are Hermes running through the Hermes Browser Extension side panel.
--- a/extension/lib/model-discovery.mjs
+++ b/extension/lib/model-discovery.mjs
@@ -422,7 +422,7 @@ export async function discoverModelsFromExternalSources({
             description: `Probed from ${clean}`,
             contextTokens: 0,
             source: 'external',
-            runtimeSelectable: true,
+            runtimeSelectable: false,
           };
         })
         .filter(Boolean);
--- a/extension/lib/model-discovery.mjs
+++ b/extension/lib/model-discovery.mjs
@@ -310,19 +310,48 @@ export function modelCatalogRefreshDecision({ previousSelectedModel = '', discov
   };
 }
 
+/**
+ * Merge model arrays from multiple sources, deduplicating by provider+rawModelId.
+ * Models sharing the same rawModelId but from different providers are kept;
+ * true duplicates (same provider + same rawModelId) are deduplicated.
+ * Earlier sources in the array order take priority (registry > dashboard > v1 > external).
+ */
+export function mergeModelsByRawId(arrays = []) {
+  const seen = new Set();
+  const out = [];
+  for (const models of arrays) {
+    if (!Array.isArray(models)) continue;
+    for (const m of models) {
+      if (!m) continue;
+      const rawId = m.rawModelId || m.id || '';
+      if (!rawId) continue;
+      const providerKey = String(m.provider || m.providerLabel || m.source || '').toLowerCase();
+      const key = `${providerKey}::${rawId.toLowerCase()}`;
+      if (seen.has(key)) continue;
+      seen.add(key);
+      out.push(m);
+    }
+  }
+  return out;
+}
+
 export function mergeModelsWithRegistry({ registryModels = [], sessionModels = [] } = {}) {
   const out = [];
   const seen = new Set();
   // Registry first (these are the gateway-blessed models)
   for (const model of registryModels) {
-    if (!model || !model.id || seen.has(model.id)) continue;
-    seen.add(model.id);
+    if (!model) continue;
+    const key = String(model.rawModelId || model.id || '').toLowerCase();
+    if (!key || seen.has(key)) continue;
+    seen.add(key);
     out.push({ ...model, source: model.source || 'registry' });
   }
   // Then session-discovered models, marking them as such
   for (const model of sessionModels) {
-    if (!model || !model.id || seen.has(model.id)) continue;
-    seen.add(model.id);
+    if (!model) continue;
+    const key = String(model.rawModelId || model.id || '').toLowerCase();
+    if (!key || seen.has(key)) continue;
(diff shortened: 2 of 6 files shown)


abundantbeing:
Verdict: Request changes
Risk: Medium

Thanks for the PR — the no-auth external probe path is a sensible start, and I verified it does not attach the Hermes bearer token to custom `/v1/models` fetches. I think this needs another pass before merge:

Blocking:
1. External endpoint selections are not actually routed to the external endpoint. `discoverModelsFromExternalSources()` builds requestable models with `provider` = hostname and `rawModelId`, while keeping the source URL only as display text (`extension/lib/model-discovery.mjs:387-402`). On send, Browser only posts `{ model: currentModelRequestId(), provider: currentModelProviderSlug() }` to the Hermes gateway/session routes (`extension/sidepanel.js:4326-4333`, `4481-4488`, `4508-4511`). A model discovered from `http://wimpy:8080/v1` therefore becomes `model=... provider=wimpy`; the custom URL is not present, so endpoints that are not already configured in Hermes cannot be used even though the picker marks them requestable. Either keep these as discovery-only/observed rows, or carry a validated source identifier through to backend support that can actually route the request.
2. Deduping only by `rawModelId` collapses distinct providers/endpoints. `mergeModelsByRawId()` uses `m.rawModelId.toLowerCase()` as the sole key (`extension/lib/model-discovery.mjs:297-313`), so two sources exposing `llama3`/`gpt-4o-mini` produce one picker row and the later provider disappears. Please dedupe on provider/source + raw model ID (or keep the prefixed `id`) so custom source choices remain distinct.

Non-blocking:
- GitHub currently reports this PR as `CONFLICTING` against `main`, so it will also need a rebase/merge refresh.
- There are no new tests covering `discoverModelsFromExternalSources()` or the multi-source merge behavior. Please add coverage for endpoint URL handling and duplicate model IDs across providers.

Verification run on `5aff7ce`:
- `npm ci` ✅
- `npm run verify` ✅ 139/139 tests passing; manifest OK
- `npm run lint` ✅ 0 errors, 5 warnings
- `npm run build` ✅
- `git diff --check origin/main...HEAD` ✅
- Added-line secret scan ✅ no common token/private-key patterns found

rahlquist:
All three review issues addressed in commit 6a4e3ba (rebased on current main):

1. **External models are now discovery-only** —  in . Models from custom endpoints appear as observed rows in the picker rather than requestable ones, since the custom endpoint URL isn't carried through to backend routing.

2. **Dedup key widened to provider+rawModelId** —  now uses a composite key . Two providers exposing a model named `phi-4` each produce distinct picker rows. True duplicates (same provider, same model) still collapse.

3. **Tests added** — 3 new subtests in common.test.mjs:
   - `mergeModelsByRawId deduplicates by provider+rawModelId composite key`
   - `discoverModelsFromExternalSources parses various response shapes`
   - `external source URL normalisation strips /v1 and appends /v1/models`

CI: 148/148 tests passing (was 139 + 6 from upstream main + 3 new). No conflicts — rebased on current main.

abundantbeing:
Thanks for the custom model-source work. The safe version landed on main in 1f026fa and remains in the current main head 931fc16: user-configured OpenAI-compatible sources are discovered via /v1/models, URL-normalized/deduped/capped, and rendered as disabled observed/discovery-only rows unless the connected Hermes runtime exposes the model. Verification on main: npm run test passed 163/163, plus lint/build passed during the release pass. Closing this PR as superseded because the branch now conflicts with main and a direct merge would drop the hardened routing guards.

PR #33, opened 2026-07-07
Security: fix parameter evasion, CWD binary hijacking, and clipboard …

Doom-pixel-alt:
## Summary
This Pull Request addresses three security risks identified in the Hermes Browser Extension codebase. All modifications have been validated, and the test suite passes successfully.

### 1. Robust URL Restriction Checking
* **Files**: `extension/lib/browser-context-protocol.mjs`, `extension/lib/common.mjs`
* **Fix**: Updated `isRestrictedUrl` to include `parsed.search` and `parsed.hash` in the evaluated URL haystack. This prevents potential privacy leaks if sensitive keywords (e.g., banking/crypto/medical keywords) are present only in the query parameters or hash segments of an active tab URL (e.g., `https://google.com/search?q=mybank`).

### 2. CWD Binary Planting Mitigation (Windows)
* **File**: `scripts/hermes-review-watch.mjs`
* **Fix**: Added a check in `githubToken` under Windows to block executing the `gh` command if a `gh` binary or script (e.g. `.exe`, `.cmd`, `.bat`, `.lnk`) exists in the current working directory (`process.cwd()`). This mitigates binary hijacking risks.

### 3. Avoid Automatic Clipboard Leakage
* **File**: `scripts/windows-setup.mjs`
* **Fix**: Disabled automatic copying of the pairing token or API key to the system clipboard during setup. Clipboard copying now requires the explicit `--clip` CLI argument. By default, secrets are securely printed to the console with a security warning.

## Verification
* Added new test cases in `tests/common.test.mjs` verifying query-parameter and hash-based URL restrictions.
* Ran `npm test` successfully; all 184 unit tests pass.

### DIFF
--- a/extension/lib/browser-context-protocol.mjs
+++ b/extension/lib/browser-context-protocol.mjs
@@ -89,7 +89,7 @@ export function isRestrictedUrl(url = '') {
     return true;
   }
   if (RESTRICTED_SCHEMES.has(parsed.protocol)) return true;
-  const haystack = `${parsed.hostname}${parsed.pathname}`;
+  const haystack = `${parsed.hostname}${parsed.pathname}${parsed.search}${parsed.hash}`;
   return SENSITIVE_URL_PATTERNS.some((pattern) => pattern.test(haystack));
 }
 
--- a/extension/lib/common.mjs
+++ b/extension/lib/common.mjs
@@ -1578,7 +1578,7 @@ export function isRestrictedUrl(url = '') {
     return true;
   }
   if (RESTRICTED_SCHEMES.has(parsed.protocol)) return true;
-  const haystack = `${parsed.hostname}${parsed.pathname}`;
+  const haystack = `${parsed.hostname}${parsed.pathname}${parsed.search}${parsed.hash}`;
   return SENSITIVE_URL_PATTERNS.some((pattern) => pattern.test(haystack));
 }
 
--- a/scripts/hermes-review-watch.mjs
+++ b/scripts/hermes-review-watch.mjs
@@ -31,6 +31,18 @@ function readEnvFileValue(name) {
 function githubToken(env = process.env) {
   if (env.GITHUB_TOKEN) return env.GITHUB_TOKEN;
   try {
+    if (process.platform === 'win32') {
+      const cwd = process.cwd();
+      const extensions = ['.exe', '.cmd', '.bat', '.lnk'];
+      for (const ext of extensions) {
+        if (fs.existsSync(path.join(cwd, `gh${ext}`))) {
+          throw new Error('Refusing to execute gh from current working directory for security reasons.');
+        }
+      }
+      if (fs.existsSync(path.join(cwd, 'gh'))) {
+        throw new Error('Refusing to execute gh from current working directory for security reasons.');
+      }
+    }
     return execFileSync('gh', ['auth', 'token'], { encoding: 'utf8', stdio: ['ignore', 'pipe', 'ignore'] }).trim();
   } catch {
     return '';
--- a/scripts/windows-setup.mjs
+++ b/scripts/windows-setup.mjs
@@ -232,15 +232,25 @@ async function main() {
 
   const pairing = await probePairing();
   let copiedFallbackKey = false;
+  const useClip = args.has('--clip');
+
   if (pairing.token) {
-    copiedFallbackKey = copySecretToClipboard(pairing.token, 'copy-pairing-token-to-clipboard');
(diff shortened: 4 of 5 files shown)


abundantbeing:
## Hermes Agent Review

## Summary

I would not merge PR #33 yet. The intent is good, but the security fixes are incomplete and `windows-setup.mjs` appears to introduce a new secret-output regression. The PR body claims `npm test` passed, but no trusted CI/check evidence was provided, so I’m not treating tests as verified.

## Blocking issues

- `isRestrictedUrl()` still has parameter/hash evasion via percent-encoding.
  - The new haystack uses raw `parsed.search` and `parsed.hash`.
  - URLs like `https://google.com/search?q=my%62ank` or `https://example.com/#%77allet` can bypass keyword regexes because `URL.search` preserves encoding.
  - Fix: include safely decoded query/hash/path variants in the haystack, with malformed-percent fallback.
  - Add tests for encoded sensitive words and malformed percent sequences.

- Windows `gh` binary hijack mitigation is incomplete.
  - The CWD check covers `gh`, `.exe`, `.cmd`, `.bat`, `.lnk`, but misses `gh.com`, which is part of the default Windows executable resolution surface and often appears before `.exe` via `PATHEXT`.
  - It may also miss other `PATHEXT`-driven entries depending on the environment.
  - Fix: resolve the actual executable path before running `gh`, then reject it if it resolves under `process.cwd()`, or require a trusted absolute `gh` path.
  - Add a regression with a fake `gh.com` in a temp cwd.

- `scripts/windows-setup.mjs --dry-run` can leak the real API key.
  - In the dry-run manual setup branch, the action is redacted.
  - But the later summary path calls `readApiServerKey()` and prints the key when `copiedFallbackKey` is false.
  - Dry-run must never read or print live secrets.
  - Fix: make summary respect `dryRun` and print only redacted/manual instructions.

- Default fallback behavior now prints long-lived `API_SERVER_KEY` to stdout.
  - This swaps clipboard leakage for terminal/log leakage.
  - For the short-lived pairing token this may be acceptable if explicitly intended, but the API key fallback is higher risk.
  - Fix: require an explicit `--show-key`/interactive confirmation or keep fallback instructions redacted by default.

## Suggestions

- Add direct coverage for both duplicated `isRestrictedUrl()` implementations, or refactor to one shared helper so `common.mjs` and `browser-context-protocol.mjs` cannot drift.
- Make `githubToken()` distinguish “security refusal” from “no token found”; silently returning `''` hides the hardening decision from operators.
- Avoid reading the API key twice in setup. Capture fallback metadata once and pass only redacted/explicitly-approved values into summary output.

## Checks to run

- `npm test`
- Focused URL tests covering:
  - raw query/hash sensitive words
  - percent-encoded sensitive words
  - malformed percent encoding
  - safe URLs with benign query/hash content
- Windows hijack regression with temp cwd files:
  - `gh.com`
  - `gh.exe`
  - `gh.cmd`
  - `gh.bat`
  - no local `gh`, valid PATH `gh`
- Setup secret-output checks:
  - `node scripts/windows-setup.mjs --dry-run` with pairing unavailable and `.hermes/.env` present
  - same without `--clip`
  - same with `--clip`
  - JSON/non-JSON output modes if supported
- Repo verifier/syntax checks, e.g. `npm run check:js` / `npm run verify` if available.

---

Reviewed signature: `a8940f97fd4c0d3c94c917b4ecc49c75c5b8582260cea78eb2bfc1adbbb0eb0a`

Doom-pixel-alt:
No problem. I hihlighted the problems and suggested some fix but if you want to make your own then you can. :)

abundantbeing:
@Doom-pixel-alt fixed on `main` in bc58ddee62b199bebad6a8c935ffa3378a300178.

Thanks for finding and reporting the restricted-URL evasion, Windows `gh` execution risk, and setup clipboard/secret-handling issues in #33.

I landed a hardened implementation rather than merging the PR branch directly:
- restricted URL detection now checks raw and decoded path/query/hash variants
- the review watcher resolves `gh` through Windows `PATH`/`PATHEXT`, blocks unsafe CWD/PATH cases, and executes the resolved binary path
- setup secret handling only copies or prints with explicit `--clip` / `--show-secret`
- clipboard copying uses the Windows System32 `clip.exe` path instead of bare `clip`
- regression tests cover the privacy/security paths

Verification run before landing:
- `npm run verify`
- `npm run build`
- `npm run lint` — 0 errors, existing warnings only
- `npm run package`
- `node scripts/check-manifest.mjs`

PR #34, opened 2026-07-10
feat: add Hermes browser action bridge

EmpireOperating:
# Hermes browser action bridge

## Summary

- add the versioned `hermes.browser.actions.v1` policy with the exact reviewed action allow-list
- route `browser.action.requested` events through an explicit side-panel approval card for `click`, `typeText`, `select`, and `openUrl`
- execute approved actions in the active tab while blocking restricted URLs, sensitive/submitting controls, unsupported actions, and non-HTTP(S) targets
- return `browser.action.result` acknowledgements with sanitized metadata only; raw screenshot `dataUrl` values remain inside the extension
- add focused policy/bridge tests and a loaded-Chromium end-to-end smoke harness whose `ws:` CSP and screenshot host-permission relaxations exist only in a temporary extension copy

## Security properties

- exact supported actions: `getSnapshot`, `screenshot`, `scroll`, `click`, `typeText`, `select`, `openUrl`
- `submitForm`, cookie/history/bookmark/native-messaging style actions, restricted URL categories, browser-internal URLs, and sensitive controls are rejected
- mutating actions always require fresh user approval; an inbound `approvedByUser` value cannot bypass the side-panel prompt
- page content is labeled untrusted
- there is no form `.submit()` execution path
- screenshot pixels are stripped before the result crosses the gateway boundary
- action events are ignored until their exact browser session is active

## Test plan

- `npm run verify`
- `npm run smoke:loaded-browser`
- `npm run lint`
- `npm run build`

The focused browser-action tests cover restricted URL/action rejection, fresh mutating-action approval, session binding, absence of a `.submit()` path, and screenshot `dataUrl` sanitization. The loaded-browser smoke verifies a real unpacked extension receives `browser.action.requested`, captures a screenshot, reports `browser.action.result`, and does not return raw screenshot data to Hermes.


abundantbeing:
## Hermes Agent Review

## Summary

This is a high-risk browser-control surface and is not ready to merge. The event provides a test plan but no successful check results, and the diff contains approval-integrity, form-submission, URL-policy, and result-correlation defects.

## Blocking issues

- Approval is not bound to the executed operation. `browserActionReceiptCopy()` may display `target.name` or `target.text`, while `browserActionTarget()` executes `target.selector` first. The approval also omits the `typeText` value. Resolve the target before prompting and bind approval to the exact action, value, tab ID, URL, frame, session, and target identity.
- Approval has active-tab/session TOCTOU vulnerabilities. The active tab is resolved only after approval, and `report()` reads the mutable current `sessionId()` after execution. A tab, navigation, or session change while the dialog is open can execute or report against the wrong context. Capture and revalidate these values immediately before execution.
- Concurrent prompts can strand requests permanently. `requestBrowserActionApproval()` removes an existing card without resolving its Promise. Serialize requests or explicitly deny displaced requests, and add timeout/disconnect/session-switch cleanup.
- The no-submit guarantee is bypassable. A `<button>` inside a form defaults to `type="submit"` even when `getAttribute('type')` is empty; `element.click()` will submit it. Check effective `element.type`, form association, and indirect scripted submit/navigation controls. The regex test for `.submit()` does not verify behavior.
- URL handling is incorrect and bypassable. `safeUrl()` reduces every accepted URL to its origin, so `openUrl("https://docs.example/path")` navigates to `https://docs.example/`. Raw substring checks also miss encoded sensitive routes such as `/%62ank` or `/my%77allet`. Preserve a full validated execution URL, create a separate redacted preview, and inspect decoded URL components.
- Result correlation is unsafe and incomplete. `{ actionType, ...result }` lets a runtime response overwrite the authoritative requested action type, while snapshot text and scroll viewport data are silently discarded by `sanitizeBrowserActionResult()`. Build the receipt explicitly, bind it to the original request/session, and define how snapshot/screenshot results are actually consumed.
- `smoke:loaded-browser` has author-specific defaults (`/home/openclaw/...` and `/home/hermes-agent/...`), manually installs the bridge, and auto-approves requests. It is not a portable fresh-clone smoke test and does not exercise production wiring or the real approval UI.

## Suggestions

- Add request-ID replay/deduplication protection.
- Consolidate the duplicated URL/action policy in `content.js` and `browser-actions.mjs` to prevent policy drift.
- Redact runtime exception messages before sending them through the gateway; browser errors may contain URLs or local details.
- Enforce the declared protocol/version rather than only adding a decorative normalized field.

## Checks to run

- `npm run verify`
- `npm run lint`
- `npm run build`
- `npm run smoke:loaded-browser` from a clean environment without author-specific paths
- Loaded-browser tests for active-tab/session/navigation changes during approval
- Concurrent prompt, timeout, disconnect, and replay tests
- Real DOM test for an implicit form-submit button
- Encoded restricted URL and benign-substring cases
- Full-path `openUrl` navigation
- Mismatched selector/display label and hidden `typeText` value
- Runtime response with a forged `actionType`
- Production side-panel startup with the real approval interaction

No approval: successful checks are not shown in the event data, and the issues above block merge.

---

Reviewed signature: `0c5c3ba99734a176d8d10b81b18b774fba50b9f3ab0a0e9c1adc75cc9d70ece6`

abundantbeing:
## Hermes Agent Review

## Summary

High release/security risk. The bridge adds useful guardrails, but it is not safe to merge as written. The event contains a test plan, not actual check results, so this review does not approve the PR.

## Blocking issues

- Approval is not bound to the action actually executed. `browserActionReceiptCopy()` displays `target.name`/`text` before `selector`, while `browserActionTarget()` executes `selector` first. A request can display “Save” but target `#delete`. The active tab, URL, session, and resolved element are also not captured and revalidated after approval, allowing a tab/session/DOM change while the prompt is open. Resolve the target before prompting and bind one-use approval to immutable session, request, tab, URL, frame, action, value, and element identity.

- Submission and navigation restrictions are bypassable. `isSensitiveOrSubmittingControl()` misses a `<button>` inside a form with no explicit `type`; its effective type is `submit`, so `element.click()` submits the form. Clicking an anchor can also navigate to a restricted destination without applying the URL policy to its `href`. Validate effective control semantics and destination URLs, and add real DOM tests.

- `openUrl` executes the wrong URL. `safeUrl()` reduces every accepted URL to `${protocol}//${host}/`, and that normalized value is passed to `chrome.tabs.update()`. An approved request for `https://docs.example/path?q=1` navigates to `https://docs.example/`. Preserve a full validated execution URL and create a separate origin-only preview value.

- URL filtering is inconsistent and bypassable. The shared policy scans the raw URL, while `content.js` scans `parsed.href`; encoded hosts/components such as `%62ank` can therefore behave differently. Raw substring matching also falsely blocks benign URLs such as `syntax.com` because it contains `tax`. Centralize policy and inspect canonicalized, decoded URL components with boundary-aware rules.

- Result/session correlation is unsafe and the result contract is unusable. `report()` reads mutable `sessionId()` after approval, so a stale action can execute and be reported under a new session. `{ actionType, ...result }` also lets the runtime overwrite the authoritative action type. Additionally, sanitization discards snapshot text, viewport data, and screenshot pixels without providing a local handle, leaving successful `getSnapshot`/`screenshot` requests with no usable result. Capture the originating session/request, construct the receipt explicitly, and define usable sanitized result payloads.

- `smoke:loaded-browser` is not portable and bypasses production behavior. It defaults to author-specific absolute paths, manually installs the bridge, and auto-approves requests, so it neither works from a normal fresh clone nor tests the real approval UI/startup path. Add dependency discovery/setup errors and a loaded-extension test using normal side-panel initialization and actual approval interaction.

## Suggestions

- Serialize approval prompts or explicitly deny displaced requests; currently removing an existing card leaves its Promise unresolved. Add timeout, disconnect, panel-close, session-switch, and replay cleanup.
- Enforce the declared `hermes.browser.actions.v1` protocol instead of unconditionally adding it during normalization.
- Redact runtime exception messages before sending `reason` through the gateway.
- Remove `BROWSER_ACTIONS_PR_BODY.md`; it duplicates the PR description and should not ship as product source.

## Checks to run

No successful checks are shown in the event data. Before reconsidering merge, run:

- `npm run verify`
- `npm run lint`
- `npm run build`
- `npm run smoke:loaded-browser` from a clean clone
- Loaded-browser regressions for mismatched display target/selector, tab and session changes during approval, concurrent prompts, implicit form submission, restricted anchor navigation, encoded URLs, full-path `openUrl`, result correlation, and real approval interaction
- `git diff --check` and an added-line secret scan

---

Reviewed signature: `665ece3186d7057790b7a4fa4cb80f24189d84f6c4545204f27a5e5e5f8baef1`

EmpireOperating:
Closing this stale draft contribution path. The Browser Companion work is continuing privately in a newer local integration; keeping the source branch intact as reference.

PR #36, opened 2026-07-11
fix(companion-plugin): handle content arrays in _last_user_message

iruzen-dono:
## The Bug

The companion plugin's `pre_llm_call` hook silently fails to detect browser context in messages that use content arrays (the OpenAI message format where `content` is a list of `{type, text}` objects instead of a plain string).

Modern Hermes versions use content arrays when attachments (images, files) are present in a message. The `_last_user_message` helper only handled `isinstance(content, str)`, so `str([{type: "text", text: "..."}])` returns `"[object Object]"` which does not contain the `UNTRUSTED_BROWSER_CONTEXT_START` marker, and the context cache remains empty.

## The Fix

1. **`companion-plugin/context_store.py`** — add `_meaningful_text_from_content_list()` helper that flattens content arrays into a single string by concatenating `.text` fields from each `type: "text"` part.

2. **`companion-plugin/hooks.py`** — update `_last_user_message()` to:
   - Check for content arrays in both `user_message` kwargs and `conversation_history` messages
   - Delegate array flattening to the new helper

### DIFF
--- a/companion-plugin/context_store.py
+++ b/companion-plugin/context_store.py
@@ -57,6 +57,24 @@ def _section_key(label: str) -> str:
     return label.strip().lower().replace(" ", "_").replace("/", "_")
 
 
+from typing import Any as _Any
+
+def _meaningful_text_from_content_list(parts: list[_Any]) -> str:
+    """Flatten a content array (OpenAI format) into a single string.
+
+    A content array looks like:
+        [{"type": "text", "text": "Hello"}, {"type": "image_url", ...}]
+    Returns the concatenated text parts for browser-context scanning.
+    """
+    chunks: list[str] = []
+    for part in parts:
+        if isinstance(part, dict) and part.get("type") == "text":
+            text = part.get("text", "")
+            if isinstance(text, str) and text.strip():
+                chunks.append(text)
+    return "\n".join(chunks)
+
+
 def _meaningful_section_text(lines: list[str] | None) -> str:
     if not lines:
         return ""
--- a/companion-plugin/context_store.py
+++ b/companion-plugin/context_store.py
@@ -8,6 +8,8 @@ Hermes restarts.
 
 from __future__ import annotations
 
+from typing import Any
+
 import hashlib
 import json
 import re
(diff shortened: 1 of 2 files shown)


abundantbeing:
## Hermes Agent Review

## Summary

No merge-blocking correctness or security issue is apparent in the supplied diff. The change safely extracts only string text from structured content parts and applies it to both direct hook input and conversation history.

I cannot approve: the event provides no CI, test, or check results.

## Blocking issues

None identified from the supplied event data.

## Suggestions

- Add focused regression tests for `_last_user_message()` covering:
  - `user_message` as a content array with text plus an image/file part.
  - `conversation_history` containing an array-based user message.
  - arrays with no text parts, malformed entries, and non-string `text` values.
  - multiple text parts, verifying marker text remains detectable.
- Move `from typing import Any as _Any` to the module’s normal import block. It is non-blocking, but avoids an import appearing mid-module.
- Consider naming the helper without the `_Any` alias unless avoiding an existing import collision is necessary.

## Checks to run

- Run the companion-plugin unit test suite, including the new structured-content regression cases.
- Run the repository’s standard lint/typecheck/test commands.
- Run `git diff --check` against the PR base.
- Verify the `pre_llm_call` hook receives browser-context markers when an attachment causes Hermes to emit OpenAI-style content arrays.

---

Reviewed signature: `b54152058879aecd4aa1a091722220e7d72e746f0f3b2cd9b580492bdb88750e`

abundantbeing:
Thanks @iruzen-dono — this is fixed and live on `main` in 50feb17. Your original commit 21491fc is retained as a merge parent, so the authored commit and contributor credit are preserved.

I added focused regression coverage for direct structured `user_message` content, structured conversation history, mixed text/image parts, malformed entries, and non-string text values. I also kept the helper with the message-extraction logic in `hooks.py`; that supersedes the late import-only cleanup in 9619766.

Verified on the integrated state:
- `npm run test`: 287/287 passed
- `npm run lint`: 0 errors (11 pre-existing warnings)
- `npm run build`: passed
- Python companion plugin compilation: passed

Closing this PR as resolved on `main` rather than merging the newer head again. Appreciate the fix and the quick follow-up.

PR #39, opened 2026-07-12
fix: redact credential-bearing URL parameters

HuntIntegrativeSolutions:
## Summary

- Treat credential-like query/hash parameters as restricted URLs before Browser Context Protocol prompt assembly.
- Apply the same guard to shared URL privacy logic in `common.mjs`.
- Add regression tests for `api_key` / `token` query/hash leakage while preserving normal public URLs.

## Problem

`v0.1.10` blocks browser internals and obvious sensitive URL categories, but credential-like query/hash parameters can still pass through tab URL context.

Example before this patch:

```text
https://example.com/docs?api_key=browser-secret-value#token=abc
```

could appear in:

- `privacySafeTabForPrompt(...)`
- `summarizeTabs(...)`
- `buildBrowserContextPrompt(...)`

## Fix

Add this sensitive URL pattern to both privacy guards:

```js
/(?:^|[?#&;\s])(?:api[_-]?key|access[_-]?token|auth[_-]?token|refresh[_-]?token|session[_-]?token|token|password|passwd|secret|private[_-]?key)=/i
```

When matched, the tab is rendered as:

```text
(restricted tab)
(omitted by privacy guard)
```

## Tests

- Adds Browser Context Protocol regression coverage for credential-like query/hash URLs.
- Adds shared `isRestrictedUrl` coverage for `?api_key=...#token=...`.
- Preserves a normal public URL control case: `?next=public` remains unrestricted.

## Verification

Ran locally:

```text
npm run verify
npm run build
```

Results:

```text
npm run verify -> PASS, 309/309 tests
Manifest OK: Hermes Browser Extension 0.1.10
npm run build -> PASS
```

Also validated with a disposable Chromium profile that prompt payloads no longer include the credential-like URL string after this patch.

## Security note

This is a privacy hardening patch for accidental credential exposure through URL metadata. It does not change browser permissions or add new data collection.

### DIFF
--- a/extension/lib/browser-context-protocol.mjs
+++ b/extension/lib/browser-context-protocol.mjs
@@ -51,6 +51,7 @@ const SENSITIVE_URL_PATTERNS = [
   /\/payments?/i,
   /\/medical|healthcare|patient|mychart/i,
   /\/tax|irs\.gov|ssa\.gov/i,
+  /(?:^|[?#&;\s])(?:api[_-]?key|access[_-]?token|auth[_-]?token|refresh[_-]?token|session[_-]?token|token|password|passwd|secret|private[_-]?key)=/i,
 ];
 
 export function clampText(value = '', maxChars = 12_000) {
--- a/extension/lib/common.mjs
+++ b/extension/lib/common.mjs
@@ -126,6 +126,7 @@ const SENSITIVE_URL_PATTERNS = [
   /\/payments?/i,
   /\/medical|healthcare|patient|mychart/i,
   /\/tax|irs\.gov|ssa\.gov/i,
+  /(?:^|[?#&;\s])(?:api[_-]?key|access[_-]?token|auth[_-]?token|refresh[_-]?token|session[_-]?token|token|password|passwd|secret|private[_-]?key)=/i,
 ];
 
 export function normalizeGatewayUrl(value = DEFAULT_SETTINGS.gatewayUrl) {
--- a/tests/browser-context-protocol.test.mjs
+++ b/tests/browser-context-protocol.test.mjs
@@ -31,7 +31,7 @@ test('buildBrowserContextPayload normalizes browser context into a stable protoc
       id: '7',
       active: true,
       title: '<img src=x onerror=alert(1)>',
-      url: 'https://example.com/private?api_key=browser-secret-value',
+      url: 'https://example.com/private/docs',
       favIconUrl: 'https://example.com/favicon.ico',
     },
     tabs: [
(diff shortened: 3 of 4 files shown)


abundantbeing:
## Hermes Agent Review

## Summary

Request changes. The patch correctly targets the reported `api_key`/`token` cases in both duplicated privacy guards, and PR #39 is currently mergeable. However, it leaves a common credential-bearing parameter unredacted. No GitHub checks are reported for head `090d6c3`; the PR body’s local verification is unverified.

## Blocking issues

- `extension/lib/browser-context-protocol.mjs:54` and `extension/lib/common.mjs:129` do not include `client_secret`, despite the existing secret-text redactor explicitly recognizing it.
  - Example: `https://example.com/docs?client_secret=value` does not match the new URL pattern or any existing sensitive-URL pattern, so it can enter browser context/prompt assembly as a full URL.
  - This conflicts with the PR’s stated credential-bearing URL hardening goal. Add `client[_-]?secret` to both guards and regression coverage for query and hash forms.

## Suggestions

- Add direct prompt-level assertions for a credential-bearing active tab, selected tab, tab summary, and receipt/payload output. The new Browser Context Protocol coverage verifies the tabs payload case, but not that the credential string is absent from every prompt-facing sink.
- Centralize the sensitive URL policy instead of maintaining two identical pattern lists. Privacy rules drifting between `common.mjs` and `browser-context-protocol.mjs` is a recurring security-regression risk.
- Expand the fixture matrix to include URL-encoded parameter names and common signed-URL credentials where in scope, e.g. `client%5Fsecret`, `x-api-key`, and signed URL signature/credential parameters.

## Checks to run

- `node --test tests/browser-context-protocol.test.mjs tests/common.test.mjs`
- `npm run verify`
- `npm run build`
- Add and run negative-leak assertions proving `client_secret` values do not appear in `buildBrowserContextPrompt(...)`, `summarizeTabs(...)`, or Browser Context Protocol receipts.

---

Reviewed signature: `dc3068a2c62c4161c02f1d03a3cfbe7efff4824e8851f828709d7ddf2d644d5d`

abundantbeing:
Thanks @HuntIntegrativeSolutions — the credential-bearing URL hardening is shipped in [v0.1.11](https://github.com/abundantbeing/hermes-browser-extension/releases/tag/v0.1.11).

The landed implementation keeps the useful direction from #39 and broadens it to one shared credential-URL policy covering query and fragment parameters, nested and encoded credentials, `client_secret`, and common signed-URL fields across every prompt-facing Browser Context Protocol sink. Verification on the released commit passed 385/385 tests plus Chromium and Firefox builds.

Your contribution is recorded in [`CONTRIBUTORS.md`](https://github.com/abundantbeing/hermes-browser-extension/blob/main/CONTRIBUTORS.md) and the GitHub-linked attribution commit [`fcb1ffe`](https://github.com/abundantbeing/hermes-browser-extension/commit/fcb1ffe9ab20445d8a9d7ab171cb9ebbfb2f8b22).

Closing this PR as resolved rather than merging because current `main` already contains the broader hardened implementation; merging this branch now would replace it with the narrower patch.

PR #40, opened 2026-07-13
chore: upgrade eslint and @eslint/js to v10

kift99:
## Summary

- Upgraded `eslint` from v9 to v10.7.0
- Upgraded `@eslint/js` from v9 to v10.0.1
- All 401 tests passing

## Changes

Updated devDependencies in `package.json` and regenerated `package-lock.json` with the latest major versions of ESLint tooling.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,8 @@
       "version": "0.1.11",
       "license": "MIT",
       "devDependencies": {
-        "@eslint/js": "^9",
-        "eslint": "^9"
+        "@eslint/js": "^10.0.1",
+        "eslint": "^10.7.0"
       },
       "engines": {
         "node": ">=20"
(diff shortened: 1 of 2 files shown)


kift99:
a

kift99:
ok

abundantbeing:
## Hermes Agent Review

## Summary

ESLint tooling is upgraded to v10 with the lockfile regenerated. The change is low surface area, but it introduces a Node.js compatibility regression not reflected in the project’s declared engine range.

## Blocking issues

- `package.json:35` still permits Node `>=20`, while ESLint 10 and its updated dependencies require Node `^20.19.0 || ^22.13.0 || >=24`.
  - A fresh install on Node 20.0–20.18 (allowed by this repo) can install dependencies but cannot reliably run linting.
  - Update the project engine constraint to at least `>=20.19.0`, and update CI/tooling documentation or version matrices if they test/support older Node 20 releases.

## Suggestions

- Add or confirm a CI matrix using the new minimum supported Node version, particularly Node 20.19.x, so future dependency upgrades cannot silently exceed the declared runtime contract.
- This diff changes only development tooling; no application-code security concern is visible in the supplied patch. Still run an install from the committed lockfile rather than relying on dependency metadata alone.

## Checks to run

- `npm ci`
- `npm run lint`
- `npm test`
- Run the above on Node 20.19.x (and any declared supported Node 22/24 versions).
- `git diff --check`

The PR body claims 401 tests pass, but the event provides no CI or command output proving that result. I would request changes until the Node engine contract is corrected and checks are reported.

---

Reviewed signature: `f1bfa6a8a622a7a49f4cb216455f1d2252bfafc397c6ceabe32459441c24b307`

abundantbeing:
Thanks for the update. We’re deferring ESLint 10 for now. This branch is based on the pre-v0.2 package state and conflicts with current main. ESLint 10 also enables new recommended rules that currently fail the project lint contract, and its supported Node range does not match the repository’s declared `>=20` range. ESLint 9 is currently audit-clean, so we won’t merge the stale lockfile. We can revisit this as a dedicated current-main tooling migration.

PR #53, opened 2026-07-28
fix: remove audioCapture and prevent duplicate context menu items

chtcrack:
Two fixes for console errors on startup:

1. **Remove `audioCapture` from optional_permissions** — Chrome only allows audioCapture for packaged apps, not extensions. Including it generates a console warning every Service Worker start.

2. **Prevent duplicate context menu registration** — `configureContextMenus()` was called from both `onInstalled` and `onStartup`. Without a guard, `contextMenus.create()` could fire before `removeAll()` finished, producing 7× "Cannot create item with duplicate id" errors.

Both errors are cosmetic (non-breaking) but noisy in the console.

### DIFF
--- a/extension/background.js
+++ b/extension/background.js
@@ -301,8 +301,12 @@ async function queueOpenSessionRequest(message, sender) {
   return { ok: true, sessionId, surface };
 }
 
+let contextMenusConfigured = false;
+
 async function configureContextMenus() {
   if (!chrome.contextMenus?.create) return;
+  if (contextMenusConfigured) return;
+  contextMenusConfigured = true;
   await chrome.contextMenus.removeAll();
   chrome.contextMenus.create({
     id: CONTEXT_MENU_ROOT_ID,
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -20,9 +20,7 @@
     "storage",
     "tabs"
   ],
-  "optional_permissions": [
-    "audioCapture"
-  ],
+  "optional_permissions": [],
   "host_permissions": [
     "http://127.0.0.1/*",
     "http://localhost/*",


abundantbeing:
## Hermes Agent Review

## Summary

Request changes. The context-menu change is low-risk, but the permission removal is incomplete and regresses voice dictation. GitHub reports no checks for head `1f506171`; reviewer-local `npm run verify` failed during tests.

## Blocking issues

- `extension/manifest.json:23` removes `audioCapture`, but `extension/voice-dictation.js:71-76,304-310` and `extension/sidepanel.js:2071-2082` still request it through `chrome.permissions`. In the visible dictation page, a rejected/false request prevents `getUserMedia()` from running, turning a startup warning fix into a voice-capture failure. Remove the `audioCapture` permission gate from both runtimes and rely on `getUserMedia()`/microphone site permission.
- `tests/common.test.mjs:1160-1169` still requires `audioCapture` in `optional_permissions`. The focused test fails at line 1167, and the standard verification path is not green. Update the contract and add coverage proving dictation reaches `getUserMedia()` without the extension permission.

## Suggestions

- Replace `contextMenusConfigured` with a shared in-flight configuration promise and clear it on failure. The current flag is set before `removeAll()` resolves, so one transient rejection suppresses all retries for that worker lifetime.
- Add a concurrency test invoking menu configuration twice before `removeAll()` settles and assert one removal plus one set of menu creations.

## Checks to run

- `npm run verify`
- `npm run build && npm run build:firefox`
- Loaded-extension Chrome smoke: startup/reload produces no permission or duplicate-menu errors.
- Voice smoke: first-use microphone prompt works and recording starts.
- Context-menu race test covering concurrent startup/installation configuration.

---

Reviewed signature: `fd6c7c842a8f3391aaea634eabf75263a972aebd6fa32d65cca6da50ac75fb2a`

abundantbeing:
Thanks for identifying both problems — the diagnoses were valid. I landed a hardened version directly on `main` in two attributed commits:

- [`b06cfff`](https://github.com/abundantbeing/hermes-browser-extension/commit/b06cfff28884ae772fae8452ed8ed77d89e92715) removes the unsupported `audioCapture` manifest/runtime gate and keeps microphone consent on the standard `getUserMedia` flow.
- [`2c6f44c`](https://github.com/abundantbeing/hermes-browser-extension/commit/2c6f44c7eac600df9af5203417ff3949e2a6e784) serializes context-menu setup with a shared in-flight promise and preserves retry behavior after failure.

Regression coverage now exercises concurrent lifecycle setup, failed-attempt recovery, and the Chromium microphone manifest/runtime contract. Verification completed with 584 tests passing, lint at 0 errors, manifest validation, and Chromium + Firefox builds.

Closing this PR as superseded by the hardened implementation on `main`, rather than merged as submitted. Thanks again for the useful report and starting point.

PR #57, opened 2026-07-31
Add user-editable context menu items in settings

### LINKED ISSUE #56: Add user-editable context menu items in settings
Currently, the right-click context menu items (Ask, Summarize, Explain, Improve, Draft Reply, Open Hermes) are hardcoded in background.js. Users cannot add, remove, reorder, toggle, or edit these items from the UI.

This feature adds full user control over context menu items from the settings panel in both the side panel and Hermes Web surfaces.

Changes:
- Move context menu item definitions to settings (chrome.storage.local)
- Add contextMenuItems, DEFAULT_CONTEXT_MENU_ITEMS, normalizeContextMenuItems, and related constants to common.mjs
- background.js reads items from settings and rebuilds menus on storage change
- Settings UI: list editor with toggle, edit label, edit prompt, reorder (up/down), add, remove
- Each item supports: enabled/disabled, title, contexts (selection/editable/page/link/image/video/audio), and either a prompt (sent to gateway) or an inline action
- CSS styled for narrow side panel with responsive flex-wrap layout
- Test updated to read from common.mjs instead of background.js for moved strings
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #56</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @kidclone3</code> (replace <code>20</code> with the amount, and <code>@kidclone3</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

kidclone3:
Resolves #56

Currently, the right-click context menu items (Ask, Summarize, Explain, Improve, Draft Reply, Open Hermes) are hardcoded in background.js. Users cannot add, remove, reorder, toggle, or edit these items from the UI.

This PR adds full user control over context menu items from the settings panel in both the side panel and Hermes Web surfaces, and fixes two real-test findings from the right-click behaviour spike that blocked the workflow.

## Part 1 — User-editable context menu items

- Move context menu item definitions to settings (chrome.storage.local)
- Add `contextMenuItems`, `DEFAULT_CONTEXT_MENU_ITEMS`, `normalizeContextMenuItems`, `INLINE_CONTEXT_ACTIONS`, `CONTEXT_MENU_CONTEXTS`, and `normalizeContextMenuItem` to common.mjs
- background.js reads items from settings and rebuilds menus on storage change via the storage.onChanged listener
- Settings UI: list editor with toggle, edit label, edit prompt, reorder (up/down), add, remove
- Each item supports: enabled/disabled, title, contexts (selection/editable/page/link/image/video/audio), and either a prompt (sent to gateway) or an inline action
- CSS uses real design tokens (--hermes-line, --hermes-ink-rgb, --hermes-muted) with responsive flex-wrap layout for narrow side panel
- Test updated to read from common.mjs instead of background.js for moved strings

Files changed: 9 (562 insertions, 15 deletions)

## Part 2 — Right-click reliability fixes (from spike)

Three commits validated against real Chrome:

### `1fb6330` — Summon the side panel inside the context-menu user gesture
The earlier builds fell back to a tab with `sidePanel.open() may only be called in response to a user gesture`. The sync open now happens inside the gesture, before any awaited API. If the sync open rejects (panel not yet enabled for the tab), the async path re-applies residency and only then opens a fallback tab.

### `71e4c9e` — Title right-click new sessions from the task at creation
Right-click new sessions kept the default "Hermes Browser Extension" name because the post-turn auto-title is gated by `autoNameSessions`, `apiKey`, and remote-dashboard rename support. Sessions are now titled at birth from the task text (`autoSessionTitleFromText`), mode-agnostic, named exactly once. The derived title also makes the post-turn auto-title logic skip (title is no longer default).

### `c76826a` — Right-click tasks attach browser context instead of forcing chat-only
Every right-click `askHermes` call set `forceChatOnly: true`, which forced `refreshContext()` to skip and nulled the browser context — the panel showed "Chat only — no browser context attached" for every right-click task, regardless of the user's context scope setting. Removing `forceChatOnly` lets `askHermes` call `refreshContext()`, which honors the user's context scope (follow-active, pinned-tab) and the control-panel Page context depth setting (`contextDepth`). Page content now arrives through the normal `browser_context` attachment — the same pipeline the composer uses.

The `contextMenuPageText` helper (added in `71e4c9e` to fold page text into the prompt body) is removed because it duplicated `refreshContext`: it captured page text but `forceChatOnly` still nulled the structured context. With `forceChatOnly` gone, `refreshContext` does the work properly.

## Testing

- 661 tests pass (`npm run test`)
- 0 lint errors (`npm run lint`) — 12 pre-existing warnings untouched
- Build succeeds (`npm run build`)
- Manual recipe in `spikes/manual-test-recipe.md` covers the live-Chrome pass

### DIFF
--- a/extension/app.css
+++ b/extension/app.css
@@ -708,5 +708,7 @@ button:disabled { cursor: not-allowed; opacity: 0.52; }
 .context-menu-mode-row { display: flex; flex-wrap: wrap; gap: 4px; align-items: center; }
 .context-menu-mode-select { flex-shrink: 0; font-size: 11px; }
 .context-menu-mode-row select { font-size: 11px; }
+.context-menu-mode-note { font-size: 10px; color: var(--hermes-muted); }
+.context-menu-editor-notice { font-size: 11px; color: var(--hermes-danger, #b3261e); margin: 0; }
 .context-menu-prompt-input { flex: 1 1 100px; min-width: 0; font-size: 11px; }
 .context-menu-editor > button { align-self: flex-start; font-size: 11px; }
--- a/extension/app.css
+++ b/extension/app.css
@@ -690,3 +690,20 @@ button:disabled { cursor: not-allowed; opacity: 0.52; }
   .error-state h2 { font-size: 36px; }
   .web-settings-dialog section { grid-template-columns: 1fr; }
 }
+
+.context-menu-editor { margin-top: 12px; display: flex; flex-direction: column; gap: 8px; }
+.context-menu-editor h4 { margin: 0; font-size: 12px; font-weight: 600; color: var(--hermes-muted, #888); text-transform: uppercase; letter-spacing: 0.05em; }
+.context-menu-items-list { display: flex; flex-direction: column; gap: 6px; }
+.context-menu-item-row { border: 1px solid var(--hermes-border, rgba(0,0,0,0.1)); border-radius: 6px; padding: 8px; display: flex; flex-direction: column; gap: 6px; background: var(--hermes-surface, rgba(0,0,0,0.03)); }
+.context-menu-item-header { display: flex; align-items: center; gap: 6px; }
+.context-menu-item-header input[type="checkbox"] { margin: 0; flex-shrink: 0; }
+.context-menu-item-title { flex: 1; min-width: 0; }
+.context-menu-item-controls { display: flex; gap: 2px; flex-shrink: 0; }
+.context-menu-item-controls button { padding: 2px 6px; font-size: 12px; min-width: 24px; }
+.context-menu-item-detail { display: flex; flex-direction: column; gap: 4px; padding-left: 22px; }
+.context-menu-item-label { font-size: 11px; font-weight: 500; color: var(--hermes-muted, #888); }
+.context-menu-contexts-checkboxes { display: flex; flex-wrap: wrap; gap: 6px; }
+.context-menu-contexts-checkboxes label { font-size: 11px; display: inline-flex; align-items: center; gap: 2px; }
+.context-menu-mode-row { display: flex; gap: 6px; align-items: center; }
+.context-menu-mode-select { flex-shrink: 0; }
+.context-menu-prompt-input { flex: 1; min-width: 0; }
--- a/extension/app.css
+++ b/extension/app.css
@@ -692,18 +692,21 @@ button:disabled { cursor: not-allowed; opacity: 0.52; }
 }
 
 .context-menu-editor { margin-top: 12px; display: flex; flex-direction: column; gap: 8px; }
-.context-menu-editor h4 { margin: 0; font-size: 12px; font-weight: 600; color: var(--hermes-muted, #888); text-transform: uppercase; letter-spacing: 0.05em; }
+.context-menu-editor h4 { margin: 0; font-size: 10px; font-weight: 600; color: var(--hermes-muted); text-transform: uppercase; letter-spacing: 0.15em; }
 .context-menu-items-list { display: flex; flex-direction: column; gap: 6px; }
-.context-menu-item-row { border: 1px solid var(--hermes-border, rgba(0,0,0,0.1)); border-radius: 6px; padding: 8px; display: flex; flex-direction: column; gap: 6px; background: var(--hermes-surface, rgba(0,0,0,0.03)); }
-.context-menu-item-header { display: flex; align-items: center; gap: 6px; }
-.context-menu-item-header input[type="checkbox"] { margin: 0; flex-shrink: 0; }
-.context-menu-item-title { flex: 1; min-width: 0; }
+.context-menu-item-row { border: 1px solid var(--hermes-line); border-radius: 0; padding: 8px; display: flex; flex-direction: column; gap: 6px; background: rgba(var(--hermes-ink-rgb), 0.035); }
+.context-menu-item-header { display: flex; align-items: center; gap: 4px; flex-wrap: wrap; }
+.context-menu-item-header input[type="checkbox"] { margin: 0; flex-shrink: 0; width: 14px; height: 14px; }
+.context-menu-item-title { flex: 1 1 120px; min-width: 0; font-size: 12px; }
 .context-menu-item-controls { display: flex; gap: 2px; flex-shrink: 0; }
-.context-menu-item-controls button { padding: 2px 6px; font-size: 12px; min-width: 24px; }
-.context-menu-item-detail { display: flex; flex-direction: column; gap: 4px; padding-left: 22px; }
-.context-menu-item-label { font-size: 11px; font-weight: 500; color: var(--hermes-muted, #888); }
-.context-menu-contexts-checkboxes { display: flex; flex-wrap: wrap; gap: 6px; }
-.context-menu-contexts-checkboxes label { font-size: 11px; display: inline-flex; align-items: center; gap: 2px; }
-.context-menu-mode-row { display: flex; gap: 6px; align-items: center; }
-.context-menu-mode-select { flex-shrink: 0; }
-.context-menu-prompt-input { flex: 1; min-width: 0; }
+.context-menu-item-controls button { padding: 2px 5px; font-size: 11px; min-width: 22px; line-height: 1.4; }
+.context-menu-item-detail { display: flex; flex-direction: column; gap: 4px; padding-left: 18px; }
+.context-menu-item-label { font-size: 9px; font-weight: 500; color: var(--hermes-muted); text-transform: uppercase; letter-spacing: 0.12em; }
+.context-menu-contexts-checkboxes { display: flex; flex-wrap: wrap; gap: 4px 6px; }
+.context-menu-contexts-checkboxes label { font-size: 10px; display: inline-flex; align-items: center; gap: 2px; color: var(--hermes-ink); }
+.context-menu-contexts-checkboxes input[type="checkbox"] { width: 12px; height: 12px; margin: 0; }
+.context-menu-mode-row { display: flex; flex-wrap: wrap; gap: 4px; align-items: center; }
+.context-menu-mode-select { flex-shrink: 0; font-size: 11px; }
+.context-menu-mode-row select { font-size: 11px; }
+.context-menu-prompt-input { flex: 1 1 100px; min-width: 0; font-size: 11px; }
+.context-menu-editor > button { align-self: flex-start; font-size: 11px; }
--- a/extension/app.html
+++ b/extension/app.html
@@ -287,6 +287,7 @@
               <h4>Menu items</h4>
               <div id="contextMenuItemsList" class="context-menu-items-list" role="list"></div>
               <button id="contextMenuAddItem" type="button" class="secondary">+ Add menu item</button>
+              <p id="contextMenuEditorNotice" class="context-menu-editor-notice" role="status" hidden></p>
             </div>
           </section>
           <section>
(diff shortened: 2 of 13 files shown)


abundantbeing:
## Hermes Agent Review

Summary

Request changes. The feature has several correctness and state-management failures that can make configured actions silently do nothing, restore deleted defaults, overwrite settings, or leave Chromium with a partial context menu. The event and live PR rollup report no checks, so this should not be approved yet.

Blocking issues

1. Unsupported context/action combinations silently no-op
   extension/background.js:381-393 requires selectionText for every prompt action, but both editors allow prompt actions on page, link, image, video, and audio contexts. Those clicks return without feedback. Inline actions are likewise exposed for contexts their field-oriented runtime may not support. Restrict combinations in the editor or implement and test every advertised mode/context payload.

2. Removing all items restores the defaults
   extension/lib/common.mjs:99-104 converts an explicit [] or an entirely invalid list back into DEFAULT_CONTEXT_MENU_ITEMS. Users therefore cannot remove every item. Clearing the final title or context also resurrects the defaults. Preserve explicit empty arrays and distinguish “setting absent” from “user chose no items.”

3. Normalization leaves stale index-based handlers
   app.js:1827-2001 and sidepanel.js:5995-6175 capture mutable array indices. Clearing a title or final context can remove an item during persistContextMenuEditorItems(), but the DOM is not rerendered. Subsequent interaction with that row can mutate the next item or access undefined. Resolve items by stable ID and rerender after any normalization that changes the list.

4. Menu rebuilds can overlap and destroy a working configuration
   extension/background.js:331-367 clears contextMenuConfigurationPromise before starting another rebuild, defeating the existing lock. Rapid storage writes can interleave removeAll() and create() calls, causing duplicate-root failures or partial menus. Serialize or coalesce rebuilds, validate the complete next configuration first, and ensure the final generation wins.

5. The two settings surfaces can overwrite each other
   Both persistContextMenuEditorItems() implementations write the entire hermesBrowserSettings object from potentially stale in-memory state. Hermes Web does not reconcile that object in its storage listener, and the side panel has no storage-change listener. Editing from both surfaces can lose menu changes or unrelated connection/model settings. Use a dedicated storage key or merge against the latest stored value with explicit onChanged reconciliation.

6. “Open Hermes Browser” has a misleading editor contract
   extension/background.js:384 special-cases the item by ID before examining its configured mode. The editor presents “Send prompt” and “Inline action,” but either selection is ignored for that ID. Represent this as an explicit open-panel action and render it accordingly, rather than dispatching special behavior from a magic ID.

Suggestions

1. Allowlist inlineAction values and reject or deduplicate duplicate item IDs and contexts.
2. Deep-clone nested contexts arrays; the fallback currently shallow-copies default items.
3. Bound menu count, title length, and prompt length before storage and menu creation.
4. Surface storage failures in the settings UI instead of logging only to the console.
5. Rebuild when contextMenuItems is deleted or reset; the truthy check at background.js:767 misses property/object removal.

Checks to run

1. Add normalization tests for missing settings, explicit [], disabled-only lists, blank titles, zero contexts, duplicate IDs, unknown inline actions, and deep-clone isolation.
2. Add DOM tests on both settings surfaces for remove-all, invalid edits, reorder followed by editing, stable-ID targeting, storage failure, and cross-surface synchronization.
3. Add delayed Chrome API tests proving rapid storage changes cannot overlap removeAll()/create() and that failed rebuilds preserve or recover the menu.
4. Add click-dispatch tests for every advertised mode/context pair and the open-panel action.
5. Run npm run verify, npm run lint, and npm run test:e2e:loaded with persistence verified after extension reload.

---

Reviewed signature: `727021e73b8aaeb73af09542e5e3dc378a313560af09b835b377e5ae359b6ff9`

abundantbeing:
## Hermes Agent Review

## Summary

The core normalization and serialized rebuild direction is solid, but this is not merge-ready. Several advertised editor and click-dispatch contracts are broken, and the event includes test code but no proof that checks passed.

## Blocking issues

- `extension/lib/common.mjs: normalizeContextMenuItem` forces any item with ID `hermes-browser-open` back into open mode. The editor lets users change that built-in to Prompt or Inline, but normalization silently discards the choice. Only explicit `open: true` or `mode: "open"` should determine behavior after migration.
- `extension/background.js: handleContextMenuClick` records only `pageUrl`. Link and media actions therefore lose the clicked `linkUrl` or `srcUrl`, so prompts such as “ingest this link/image” receive the containing page instead. Capture, sanitize, store, and consume the context-specific target.
- Prompt limits disagree: normalization accepts 8,000 characters, while `extension/sidepanel.js: contextMenuPromptText` truncates them to 500. The editor silently saves content that will not reach Hermes. Use one shared limit and test the exact consumer-visible prompt.
- Both `persistContextMenuEditorItems` implementations issue unsynchronized read-modify-write operations. Rapid edits can complete out of order, causing an older edit to overwrite the latest one and storage events to roll back the UI. Serialize or generation-guard writes.
- The app and side-panel storage listeners do not reset when `hermesBrowserSettings` remains an object but its `contextMenuItems` property is deleted. Background resets to defaults while an open editor retains stale items. Handle property deletion consistently.
- Empty or disabled-only configurations still create the root menu with no usable children. Clicking it has no dispatch behavior. Skip root creation when no items are enabled.
- `allowFallback` changed from `false` to `true` for context-menu actions while both settings surfaces still promise that Open Hermes uses the side panel only. This can regress into opening an extension tab. Restore the side-panel-only behavior or explicitly change the product contract and test it.
- The rebuild serialization test does not actually generate concurrent changes: it waits 35 ms between triggers while the delayed operation lasts 20 ms. Fire all changes before awaiting settlement so the test would fail without serialization.

## Suggestions

- Bound and validate IDs, including rejecting collisions with `hermes-browser-root`.
- Deep-clone `contexts` in editor snapshots to prevent shared mutable state between `settings` and the draft.
- Remove or use the newly imported `DEFAULT_SETTINGS` and `normalizeContextMenuItem` symbols if lint reports them as unused.
- Add visible validation instead of silently deleting an item when its final title or context is cleared.

## Checks to run

- `node --test tests/context-menu.test.mjs tests/browser-runtime.test.mjs`
- Full repository test and lint commands.
- Add tests for built-in mode conversion, root-ID collision, childless-root behavior, property deletion, out-of-order storage completion, and the full 8,000-character prompt contract.
- Add click-dispatch tests for selection, page, link, image, video, audio, editable, inline, and open modes, asserting `selectionText`, `pageUrl`, `linkUrl`, and `srcUrl`.
- Run loaded-extension E2E on both settings surfaces, including remove-all, disable-all, reorder, rapid edits, persistence after reload, and side-panel fallback behavior.

Do not approve yet. No successful checks are shown in the event data.

---

Reviewed signature: `69fa733bccdcef032803d9a9a996b8ef6a3baeeb88b331637fb36cee0304075d`

kidclone3:
I have debug this by hand and fix the bug of opening side panel. Now it works.

https://github.com/user-attachments/assets/bc65879b-403c-459b-ac5d-5eeca41bcfc4

kidclone3:
Wait, it still has some errors, i will update soon

abundantbeing:
## Hermes Agent Review

## Summary

Request changes. PR #57 is mechanically mergeable at `c76826a`, but it is not release-ready. Multiple correctness and privacy issues remain, and GitHub reports no checks. The test claims in the PR body are not verification evidence.

## Blocking issues

1. **Custom actions fail after an MV3 worker restart**
   - `cachedContextMenuItems` starts with defaults and is refreshed only through install/startup or storage-change paths.
   - Chrome can wake a suspended worker directly for a persisted context-menu click without firing `onStartup`.
   - A custom ID then misses the cache and `handleContextMenuClick` silently returns. Edited built-ins can also use stale behavior.
   - Hydrate storage on every worker instance or re-read settings on cache miss. Add a cold-worker regression test.

2. **Actions are not bound to the clicked target**
   - The request stores `tabId` and `pageUrl`, but drops `frameId`, `linkUrl`, and `srcUrl`.
   - Browser context is later resolved from the active/pinned scope rather than the stored clicked tab.
   - Background-route tasks receive text only, so page/link/media context differs by route.
   - Capture and consume the exact tab, frame, and resource URL consistently across all routes. Do not attach full-page content to selection-only actions unless explicitly requested.

3. **Saved configuration differs from executed configuration**
   - `normalizeContextMenuItem` forces ID `hermes-browser-open` into open mode even when the editor selects Prompt or Inline.
   - The editor accepts prompts up to 8,000 characters, while `contextMenuPromptText` silently truncates them to 500.
   - Use an explicit action type instead of magic-ID behavior and one shared prompt limit.

4. **Rapid edits can overwrite newer settings**
   - Both settings surfaces run independent asynchronous `storage.get()` plus `storage.set()` operations with no serialization or generation guard.
   - Older writes can finish last and overwrite newer edits.
   - Property-only deletion of `contextMenuItems` also leaves open editor surfaces stale while the background resets to defaults.
   - Serialize writes and reconcile missing properties consistently.

5. **Panel opening can silently lose the action**
   - The lenient path treats a resolved `sidePanel.open()` promise as proof that a visible panel opened.
   - A resolved no-op leaves the stored request without a visible consumer. A rejection instead opens a fallback tab despite UI copy promising side-panel-only behavior.
   - Preserve visibility confirmation with an idempotent fallback, or change the documented contract.

## Suggestions

- Skip root-menu creation when there are no enabled children.
- Reject reserved IDs such as `hermes-browser-root`.
- Extract the duplicated editor implementation into a shared module.
- Replace source-regex tests with behavioral coverage where possible.

## Checks to run

- Cold-worker dispatch for custom and edited built-in items.
- Full context × mode × route matrix with exact tab/frame/resource assertions.
- Selection privacy and clicked-tab-versus-pinned-tab tests.
- Out-of-order storage-write tests across both settings surfaces.
- Confirmed, rejected, and resolved-but-not-visible panel-open tests.
- Loaded-extension Chrome E2E covering persistence after worker suspension.
- Reported GitHub checks for tests, lint, and production build. No checks are currently shown, so this PR should not be approved.

---

Reviewed signature: `47c0ee03edfcdc6dbe7aea275d45b28e8cc658bd218b015b09738b1c48561148`

abundantbeing:
Integrated and hardened into `main`.

Final commits:
- `7d77a7a` implementation
- `509fca1` ancestry-preserving merge

The shipped implementation keeps @kidclone3's contribution history while resolving the review blockers with:
- a shared settings editor for Hermes Browser and Hermes Web
- MV3 cold-worker-safe listener registration and request persistence
- exact tab, frame, window, menu revision, and resource binding
- serialized configuration changes with rollback to the last good menu
- independent rapid-click queueing and correct surface claiming
- restricted-page and stale-target rejection
- explicit global locale fallback entries across all 21 supported locales

Final verification on current `main`:
- `npm run test`: 685/685 passed
- `npm run lint`: 0 errors, 10 existing warnings
- Chromium build passed
- Firefox build passed
- i18n: 1,119 messages, 21 locales, 4 surfaces
- MV3 cold-worker probe passed
- `git diff --check` passed
- staged secret scan: 0 findings
- browser-rendered editor and accordion interaction verified

The loaded-extension E2E remains blocked by this Windows Chrome host exposing a page without `chrome.runtime`; the exact same failure was reproduced from a clean detached pre-integration `origin/main` worktree at `tests/e2e-loaded-extension.mjs:405`, with zero mock requests. The deterministic built-worker probe passes.

Thanks @kidclone3 for the original contribution and follow-up fixes.

PR #58, opened 2026-08-01
Add MiniMax-M2.7 context fallback

octo-patch:
Reason: Add explicit MiniMax-M2.7 context fallback metadata so model discovery reports its 204800-token window without relying on the provider catch-all.

- Add explicit hyphenated and slash-form fallback aliases for MiniMax-M2.7.
- Extend the context fallback regression test to cover the model and its explicit entries.

Checks:
- `node --test tests/common.test.mjs`
- `PYTHON=python3 npm run verify`
- `npm run lint`

### DIFF
--- a/extension/lib/common.mjs
+++ b/extension/lib/common.mjs
@@ -1229,6 +1229,8 @@ const MODEL_CONTEXT_FALLBACKS = Object.freeze([
   ['qwen', 131_072],
   ['minimax-m3', 1_000_000],
   ['minimax/m3', 1_000_000],
+  ['minimax-m2.7', 204_800],
+  ['minimax/m2.7', 204_800],
   ['minimax', 204_800],
   ['glm-5.2', 1_048_576],
   ['glm', 202_752],
--- a/tests/common.test.mjs
+++ b/tests/common.test.mjs
@@ -1608,9 +1608,16 @@ test('normalizeHermesModels does not keep default hermes-agent fallback when rea
   assert.deepEqual(models.map((model) => model.id), ['openai-codex:gpt-5.5']);
 });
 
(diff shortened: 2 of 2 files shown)


abundantbeing:
## Hermes Agent Review

## Summary

Small, low-risk metadata change. The explicit MiniMax-M2.7 aliases precede the generic `minimax` fallback and preserve the expected 204,800-token value. No security, permission, or runtime-routing regression is apparent in the diff.

## Blocking issues

- Verification evidence is missing. GitHub reports an empty `statusCheckRollup`, so the commands listed in the PR body are unverified claims. Do not approve or merge until the required checks report success.

## Suggestions

- `tests/common.test.mjs:1611-1620` behaviorally tests only `minimax:MiniMax-M2.7`. The two source-regex assertions prove that aliases exist, but not that each alias resolves correctly. Add `normalizeHermesModels` cases for slash-form input such as `minimax/minimax-m2.7`, and ideally `MiniMax-M2.7-highspeed`.
- Consider avoiding implementation-coupled source assertions if direct behavioral coverage can exercise both aliases.

## Checks to run

- `node --test tests/common.test.mjs`
- `PYTHON=python3 npm run verify`
- `npm run lint`
- `git diff --check`

No approval is recommended until those checks are actually reported.

---

Reviewed signature: `86312bc520479438bcb0336ea4fea7e159948c8898fcddc753db608759c4ba15`

abundantbeing:
Thanks @octo-patch — current main already resolves MiniMax-M2.7 to the same 204,800-token context fallback through the generic minimax entry, verified live against normalizeHermesModels, so this change would add no runtime behavior. Closing as superseded by main; the fallback test coverage remains in tests/common.test.mjs.

PR #59, opened 2026-08-03
fix: recover dashboard transport after clearing API token

DMelisena:
## Summary
- switch a remote connection from remote-api to remote-dashboard when its stored token is cleared
- initialize dashboard WebSocket during startup readiness instead of treating a blank token as unconfigured
- add regression coverage for transport selection after token removal

## Root cause
Clear stored token removed the API key but retained gatewayMode=remote-api. The next probe therefore emitted Remote Hermes API is not reachable for a dashboard URL and never attempted ws-ticket/WebSocket transport.

## Verification
- npm run verify: 184 tests passed
- npm run package: extension archive built

### DIFF
--- a/extension/lib/common.mjs
+++ b/extension/lib/common.mjs
@@ -132,6 +132,10 @@ export function normalizeGatewayMode(value = DEFAULT_SETTINGS.gatewayMode) {
   return GATEWAY_MODES.some((mode) => mode.value === normalized) ? normalized : DEFAULT_SETTINGS.gatewayMode;
 }
 
+export function gatewayModeAfterTokenClear(value = DEFAULT_SETTINGS.gatewayMode) {
+  return normalizeGatewayMode(value) === 'local-api' ? 'local-api' : 'remote-dashboard';
+}
+
 export function normalizeSessionStartupMode(value = DEFAULT_SETTINGS.sessionStartupMode) {
   const normalized = String(value || DEFAULT_SETTINGS.sessionStartupMode).trim().toLowerCase();
   return normalized === 'resume-last' ? 'resume-last' : 'new-session';
--- a/extension/sidepanel.js
+++ b/extension/sidepanel.js
@@ -26,6 +26,7 @@ import {
   escapeHtml,
   extractAssistantText,
   formatUpdateStatus,
+  gatewayModeAfterTokenClear,
   gatewayConnectionTroubleshooting,
   gatewayConnectionSummary,
   groupModelsForMenu,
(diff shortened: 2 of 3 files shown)


abundantbeing:
## Hermes Agent Review

Summary

The transport-mapping helper is reasonable, but this PR is not merge-ready. The live PR is conflicting, has no reported GitHub checks, and the recovery flow remains incomplete.

Blocking issues

1. extension/sidepanel.js:5263 and 4852-4912: the primary Connect button still calls connectToHermes(), which always uses REST health and pairing endpoints. After clearStoredToken() switches to remote-dashboard, clicking Connect can repeat the original failure instead of calling ensureRemoteWsClient(). Branch the connect action by transport and test the same-panel clear → reconnect flow.

2. extension/sidepanel.js:1016-1024 and 5785-5787: remote-dashboard startup opens the WebSocket, but initializeSessionForPanelOpen() immediately returns because apiKey is blank. Readiness then marks sessionBinding ready using the persisted settings.sessionId without creating or resuming a session on the current socket. Allow keyless WS initialization and require a confirmed WS session ID before publishing ready.

3. extension/sidepanel.js:4839-4893: clearing the token does not invalidate an in-flight pairing attempt. A pending pollPairing() can later assign a new token and persist it, undoing the user’s deletion. Add an abort/generation guard invalidated by clearStoredToken().

4. GitHub reports mergeable: CONFLICTING and mergeStateStatus: DIRTY. Conflicts must be resolved before merge.

Suggestions

Use a neutral connection status after clearing the token. setStatus('ok', ...) currently risks implying that the dashboard connection succeeded before any ticket or WebSocket handshake completes.

Checks to run

Resolve conflicts, then run npm run verify and npm run package in CI. No GitHub checks are currently reported, so the PR-body verification claim is not approval evidence.

Add behavioral coverage for clear → storage reload → remote-dashboard dispatch, clear → primary Connect → WebSocket connection, confirmed session binding before readiness, local-api preservation, invalid URL and signed-out dashboard failures, and clearing during an active pairing attempt.

---

Reviewed signature: `34f94fbc5babda99f6211b4516de83837c19008438d420d590b13844f9d83231`

abundantbeing:
Thanks @DMelisena — the transport-recovery intent from this PR landed on main in 231b20e with attribution. Clearing a remote API token now switches the connection from the REST remote-api path to the ticket-based remote-dashboard WebSocket and re-runs startup readiness, so the dashboard attach completes without manual reconfiguration. The original branch predated the schema-v1 connection-settings model, so the fix was integrated against current main rather than merged as-is. Closing as superseded.

PR #63, opened 2026-08-05
fix(delegation): surface subagent batch results after the turn stream ends

### LINKED ISSUE #55: Side panel does not show subagent status or results after delegate_task dispatch
## Summary

When the main agent dispatches a `delegate_task` call during a chat turn, the side panel shows the assistant's "Dispatched" message and then goes idle. The subagent runs in the background and its result is injected back into the session by the Hermes server, but the side panel never surfaces this — no status indicator, no result message, no automatic refresh. The user only sees the subagent's work if they manually send another message (which triggers a history reload) or switch sessions.

## Reproduction

1. Open the side panel, connect to a Hermes gateway
2. Ask the agent to do something that triggers `delegate_task` (e.g. "delegate the wiki ingestion to a subagent")
3. Observe: the assistant says "Dispatched to the subagent" and the turn completes normally
4. Wait for the subagent to finish (1–3 minutes)
5. **Nothing appears in the side panel** — nospinner, no "subagent running" indicator, no result

## Root cause

The side panel's chat rendering is built on a request-response SSE stream:
- `streamSessionChat()` sends the turn, streams the response, and calls `onDelta`/`onTool` callbacks during the stream
- When the assistant emits "Dispatched" and `run.completed` fires, the stream ends — the turn is done
- The subagent then runs **outside that stream**. Its result arrives as a server-injected user message (`[ASYNC DELEGATION BATCH COMPLETE — deleg_XXX]`) followed by a new assistant response
- The extension has **no listener** for delegation completion and **no polling loop** to detect new messages after the stream closes
- `loadSessionMessages()` (sidepanel.js:4977) only runs when: a session is opened, the refresh button is clicked, or a new user message is sent
- `renderMessagesFromStorage()` (sidepanel.js:5736) renders from a local `messages[]` array that is not updated between turns
- The task-stack widget (`task-stack.mjs`) only tracks `todo` tool events, not `delegate_task` dispatches

## Proposed fix

Two options:

**Option A (simpler):** After detecting `delegate_task` in an SSE `tool.*` event, start a polling loop that calls `GET /api/sessions/{id}/messages` at intervals (e.g. every 3s, timeout after 5min) until the delegation result message appears, then reload the message list.

**Option B (correct):** Maintain a persistent WebSocket connection (the dashboard-bridge transport already does this in remote mode) that receives async delegation-complete push events. On receipt, reload the message list and surface a toast notification.

Either way, the panel should also show a "subagent running" indicator after the dispatch turn ends, similar to the existing `tool-activity` indicator but persistent across the turn boundary.

## Affected files

- `extension/sidepanel.js` — `streamSessionChat()`, `readSseResponse()`, `loadSessionMessages()`, `renderMessagesFromStorage()`
- `extension/lib/task-stack.mjs` — does not recognise `delegate_task` events
- `extension/lib/runtime-events.mjs` — `normalizeBrowserRuntimeEvent` does not classify delegation completions
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #55</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @kidclone3</code> (replace <code>20</code> with the amount, and <code>@kidclone3</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

iruzen-dono:
## The Bug

Reported in #55: when the Hermes agent dispatches a `delegate_task`, the turn SSE stream ends with the dispatcher's acknowledgement, and the subagent batch result is injected into the session **later**, asynchronously. The side panel never renders it — no spinner, no result — until the user manually sends another message, which finally triggers a session reload.

## The Fix

A post-turn **delegation watch** in the side panel:

1. **Detection** — `captureTaskToolEvent` now flags any delegate/subagent tool event, and the turn's final answer is checked for an async-delegation marker (fallback for gateways that acknowledge dispatch in the stream text).
2. **Polling** — after a turn that triggered a delegation, the panel polls `loadSessionMessages()` every 2.5s (works for both REST local-api and remote-dashboard WebSocket modes) until the batch lands.
3. **Settling** — the watch stops when the session grew past the turn baseline, when the last assistant message carries the gateway's `[ASYNC DELEGATION BATCH COMPLETE — deleg_XXX]` marker, or after a hard 10-minute timeout. The injected result renders automatically via the existing session render path.
4. **Hygiene** — a new user turn or session switch cancels the watch; polling pauses while a send is in flight.

New pure module `extension/lib/async-delegation.mjs` keeps the logic unit-testable (`isDelegationToolEvent`, `turnMentionsDelegation`, `delegationBatchSettled`).

## Tests

- 7 new unit tests in `tests/async-delegation.test.mjs` covering event-name shapes, answer markers, session-growth settling, completion-marker settling, timeout behavior.

## Verification

- `node --test tests/*.test.mjs` → **709/709 pass** (the single failing test, `content-extraction-runtime`, is a pre-existing Windows CRLF artifact: `content-extractor.js` is checked out with CRLF while the generator emits LF — it fails identically on a pristine `origin/main` checkout with no changes; CI on Linux is unaffected).
- `npm run lint` → 0 errors (10 pre-existing warnings in sidepanel.js, untouched).
- `node --check` on both changed JS files passes.

Fixes #55

### DIFF
--- a/extension/lib/async-delegation.mjs
+++ b/extension/lib/async-delegation.mjs
@@ -1,0 +1,45 @@
+// Async delegation helpers for the browser surfaces.
+//
+// The gateway dispatches delegate_task asynchronously: the turn stream ends
+// with the dispatcher's acknowledgement and the subagent batch result is
+// injected into the session later as a separate message. These helpers decide
+// when a turn triggered a delegation and when the injected batch has settled,
+// so the side panel knows to poll the session until the result lands.
+
+const ASYNC_DELEGATION_BATCH_MARKER = /\[ASYNC DELEGATION BATCH COMPLETE|deleg_[a-z0-9_-]+/i;
+const DELEGATION_TURN_MARKER = /\[ASYNC DELEGATION|deleg_[a-z0-9_-]+|dispatched.*subagent|subagent.*dispatched/i;
+
+const DEFAULT_POLL_MAX_MS = 10 * 60 * 1000;
+
+/**
+ * Whether a streamed tool event looks like a delegate/subagent dispatch.
+ * The gateway emits tool.* and hermes.tool.progress events with a raw
+ * toolName; both names are checked so remote-dashboard and local-api
+ * surfaces behave identically.
+ */
+export function isDelegationToolEvent(event = {}) {
+  const toolName = String(event?.toolName || event?.name || event?.rawName || '').toLowerCase();
+  return toolName.includes('delegate') || toolName.includes('subagent');
+}
+
+/**
+ * Whether the turn's final answer carries an async-delegation marker. This is
+ * a fallback for gateways that acknowledge a dispatch in the stream text
+ * without a separate tool event (e.g. "[ASYNC DELEGATION — deleg_abc]").
+ */
+export function turnMentionsDelegation(finalAnswer = '') {
+  return DELEGATION_TURN_MARKER.test(finalAnswer || '');
+}
+
+/**
+ * Whether an injected subagent batch has landed: either the session grew past
+ * the turn baseline (the batch result was appended) or the last assistant
+ * message carries the gateway's async-delegation completion marker. A hard
+ * time cap stops the watch from running forever if the gateway never reports.
+ */
+export function delegationBatchSettled(messages = [], baselineCount = 0, startedAt = 0, maxMs = DEFAULT_POLL_MAX_MS) {
+  if (!Array.isArray(messages) || messages.length > baselineCount) return true;
+  const last = messages.length ? messages[messages.length - 1] : null;
+  if (last?.role === 'assistant' && ASYNC_DELEGATION_BATCH_MARKER.test(String(last.content || ''))) return true;
+  return Date.now() - startedAt >= maxMs;
+}
--- a/extension/sidepanel.js
+++ b/extension/sidepanel.js
@@ -154,6 +154,7 @@ import {
 } from './lib/capabilities.mjs';
 import { normalizeBrowserRuntimeEvent, reduceAssistantStreamText } from './lib/runtime-events.mjs';
 import { taskStackFromToolEvent, taskStackProgress, updateTaskStackStore } from './lib/task-stack.mjs';
+import { isDelegationToolEvent, turnMentionsDelegation, delegationBatchSettled } from './lib/async-delegation.mjs';
 import { classifyTurnRecovery, latestAssistantAfterUser, sessionContextFailureRecovery } from './lib/turn-recovery.mjs';
 import { createDiffusionCanvas, diffusionVariantForSeed } from './lib/diffusion-canvas.mjs';
 import { buildSupportDiagnostics } from './lib/support-diagnostics.mjs';
@@ -567,6 +568,7 @@ function renderTaskStack() {
 }
 
 async function captureTaskToolEvent(event) {
+  if (isDelegationToolEvent(event)) sawDelegationToolThisTurn = true;
   const tasks = taskStackFromToolEvent(event);
   if (!tasks) return false;
   const sessionId = String(settings.sessionId || '').trim();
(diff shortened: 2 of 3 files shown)


abundantbeing:
## Hermes Agent Review

Summary

Merge should be blocked. The watcher can immediately misclassify the original dispatch acknowledgement, timeouts, and history-load failures as successful delegation completion. GitHub reports no checks for the current head, so the PR body’s test claims are not independently verified.

Blocking issues

1. extension/lib/async-delegation.mjs: ASYNC_DELEGATION_BATCH_MARKER matches any bare deleg_* ID. The original acknowledgement normally contains that ID, so unchanged history settles on the first poll. A direct probe returned true for “[ASYNC DELEGATION — deleg_abc123] Dispatched.” Require the exact batch-complete marker rather than accepting an ID alone.

2. delegationBatchSettled returns true on timeout, and sidepanel.js renders every true result as “Delegation complete / Subagent results loaded.” A timeout must produce a distinct timed_out outcome with retry guidance, not false success.

3. loadSessionMessages is not safe for polling. On failure it appends a visible system message; that increases messages.length beyond baseline, and the watcher ignores the loader’s false return, incorrectly reporting completion. Use a quiet, side-effect-free fetch and settle only after an accepted response containing an authoritative result.

4. Session switching is race-prone. The watcher calls loadSessionMessages without a request token, and the loader can replace messages, persist them, and render before the watcher’s post-await generation check. A session switch during an in-flight poll can therefore write the old session into the new session UI/storage. Pass a cancellation token through the loader and validate it before every mutation, save, and render.

Suggestions

- Start watches from a structured, successful delegate_task dispatch with a delegation ID; broad assistant-prose matching can trigger ten minutes of unnecessary polling.
- Support and test the actual transport field shapes, including tool_name if emitted.
- Represent pending, completed, timed_out, cancelled, and load_failed explicitly rather than with one boolean.

Checks to run

- Add tests proving an unchanged acknowledgement containing deleg_* remains pending.
- Test timeout and transient load failure UI states.
- Test a deferred poll response released after switching sessions.
- Exercise local REST and remote WebSocket history paths separately.
- Run node --test tests/async-delegation.test.mjs
- Run node --test tests/*.test.mjs
- Run npm run lint
- Run node --check on both changed JavaScript modules

---

Reviewed signature: `9d297eb8b70242d490c996885fdc12b8f762fecef0792f7be26c5ebd98d5ae5c`

kidclone3:
I have tried this solution before, but then I must abandon it because the suggested solution from LLM doesn't work.

abundantbeing:
Integrated on `main` via merge commit `5bc7063`, preserving contributor commit `898a0cf` as the second parent.

The implementation was hardened before merge with one shared exact-ID watcher across the side panel and Hermes Web, bounded durable watch state, REST and Dashboard WS history reconciliation, session-switch race guards, timeout/error distinction, and duplicate suppression.

Verified before merge:
- 743/743 tests
- Chromium and Firefox builds
- 21 locale packs across 4 extension surfaces
- loaded-extension E2E on side panel and Hermes Web
- live Comet delegation returning automatically without another prompt or refresh
- no duplicate render and no visible internal completion marker

Thanks for the contribution.

abundantbeing:
HBE - PR 63 Delegation Result Return Fix Update AD

abundantbeing:
Delayed read-only review follow-up landed on `main` in `589bd21`.

Hardening added:
- storage-write failures no longer stop the delegation poll chain
- timed-out watches receive a fresh bounded poll budget when resumed
- CRLF completion markers are recognized and remain hidden from transcripts
- degraded REST stream, non-stream fallback, and accepted-turn recovery paths rescan canonical session history for structured delegation dispatches
- loaded-extension E2E now uses distinct side-panel and Hermes Web session IDs with independent counters and cross-session negative assertions

Verified on the follow-up tree: 747/747 tests, JavaScript checks, lint with 0 errors, Chromium and Firefox builds, 1,119 i18n messages across 21 locales and 4 surfaces, isolated loaded-extension E2E, 9/9 adversarial watcher probe, GitNexus refresh with 0 cycles, and clean diff checks.

PR #66, opened 2026-08-07
feat(cloud): opt-in browser-context sharing for Cloud & dashboard gateways (Phase 0)

abundantbeing:
## Summary

Unlocks the first step toward full browser-context parity for **Hermes Cloud** and **no-key Remote (dashboard WebSocket)** connections: an explicit, per-connection **consent setting** that lifts the hard Chat-only enforcement when the user opts in.

Today, `contextScopeForGateway()` hard-forces `chat-only` for `remote-dashboard` gateway mode (Cloud and ticketed dashboard transports). BCP v2 turn envelopes ride inside the user-message text, so they already flow over `prompt.submit` identically to REST — the only blocker is the scope gate. With this PR, enabling **"Share page context with this gateway"** lets Cloud/dashboard users use Follow active tab, Pin current tab, and per-tab prompt selection, with the full BCP v2 payload delivered over the dashboard socket.

## What changed

- **`shareBrowserContext` setting** (`DEFAULT_SETTINGS`, defaults)
  - **Hermes Cloud → OFF** by default (third-party endpoint; sharing is opt-in by design)
  - **Self-hosted remote dashboard → ON** by default (user's own server)
  - Local/Remote API transports unaffected (always share; gate doesn't apply)
- **`contextScopeForGateway()`** only forces Chat-only for dashboard gateways when consent is absent — all three call sites pass the flag (startup normalization, `applyContextScope`, turn-time scope resolution)
- **Settings UI**: toggle card rendered only for Cloud/dashboard connections; `app.js` context labels show "Tab context shared" vs "Chat only"
- **Capability synthesis**: remote-dashboard reports `browserContextInline` when consent is on
- **Tests**: consent gating (`tests/context-scope.test.mjs`), per-connection defaults + legacy migration (`tests/connection-modes.test.mjs`), updated scope-policy guard (`tests/common.test.mjs`)

## Test results

`node --test tests/*.test.mjs` → **751 pass / 9 fail** — the 9 failures are pre-existing on `main` in this environment (firefox-build, i18n-runtime, inline-draft-policy, site-adapters, context-menu-editor, content-extraction-core, assistant-clipboard); verified identical set on a clean checkout.

## Design notes / follow-ups

- Consent defaults: Cloud OFF / self-hosted dashboard ON — deliberate (third-party vs. own-server trust model).
- This is **Phase 0** of the browser-context-parity plan (Phases 1–6: full inline delivery verification, on-demand `browser.context.upload` RPC, live events, browser control, file delivery, docs).
- Future phases need a small Hermes-core WS dispatcher extension point (`handle_ws_method`) for the out-of-band RPCs — spec'd in the plan.

### DIFF
--- a/extension/app.js
+++ b/extension/app.js
@@ -692,9 +692,13 @@ function renderConnectionTruth({ status = 'idle' } = {}) {
   els.modelLabel.textContent = model;
   els.profileLabel.textContent = profile;
   els.connectionDot.className = `truth-dot ${status === 'online' ? 'online' : status === 'error' ? 'error' : ''}`.trim();
-  els.contextMode.textContent = translateUiText(mode === 'cloud' || settings.connectionTransport === 'remote-dashboard' ? 'Chat only' : 'Inherited safely');
+  els.contextMode.textContent = translateUiText(mode === 'cloud' || settings.connectionTransport === 'remote-dashboard'
+    ? (settings.shareBrowserContext === true ? 'Tab context shared' : 'Chat only')
+    : 'Inherited safely');
   els.contextSource.textContent = mode === 'cloud' || settings.connectionTransport === 'remote-dashboard'
-    ? translateUiText('Cloud/dashboard context disabled')
+    ? (settings.shareBrowserContext === true
+      ? translateUiText('Browser context attached to this gateway')
+      : translateUiText('Cloud/dashboard context disabled'))
     : handoff.sourceTabId ? t('context.browser_tab_handoff', { tabId: handoff.sourceTabId }) : translateUiText('No browser context attached');
   els.diagConnection.textContent = `${label} · ${settings.connectionTransport || settings.gatewayMode || translateUiText('unknown transport')}`;
   els.diagGateway.textContent = gatewayOrigin(settings.gatewayUrl);
--- a/extension/lib/capabilities.mjs
+++ b/extension/lib/capabilities.mjs
@@ -51,6 +51,7 @@ export const DEFAULT_GATEWAY_CAPABILITIES = Object.freeze({
   pluginActions: false,
   approvalEvents: false,
   browserControl: false,
+  browserContextInline: false,
   dashboardWs: false,
   endpoints: DEFAULT_ENDPOINTS,
   raw: null,
--- a/extension/lib/common.mjs
+++ b/extension/lib/common.mjs
@@ -56,6 +56,7 @@ export const DEFAULT_SETTINGS = Object.freeze({
   apiKey: '',
   tokenSource: '',
   lastConnectionTestedAt: 0,
+  shareBrowserContext: false,
   sessionId: 'hermes-browser-extension',
   sessionTitle: 'Hermes Browser Extension',
   sessionSource: 'hermes_browser',
--- a/extension/lib/connection-modes.mjs
+++ b/extension/lib/connection-modes.mjs
@@ -42,6 +42,17 @@ export function transportUsesDashboardTicket(value = '') {
   return TICKET_TRANSPORTS.has(normalizeConnectionTransport(value));
 }
 
+// Consent default for sharing browser context with a gateway. Cloud is a
+// third-party endpoint, so it stays OFF unless the user opts in. A self-hosted
+// remote dashboard is the user's own server, so it defaults ON. Local/Remote
+// API transports always share context; the value is unused by the scope gate.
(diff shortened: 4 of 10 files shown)


abundantbeing:
## Hermes Agent Review

## Summary

This introduces a useful consent gate, but the implementation does not yet guarantee Cloud opt-in. No approval: the provided test report is author-supplied and still reports 9 failures, with no successful CI checks shown.

## Blocking issues

1. Cloud can inherit sharing enabled from another connection
   - `shareBrowserContext` is one global setting rather than a per-connection choice.
   - A self-hosted dashboard can set/default it to `true`; switching that configuration to Cloud leaves it `true`, causing page context to be shared without explicit Cloud consent.
   - Reset or derive consent when the connection identity/type changes, or store consent by connection. Cloud must remain `false` until explicitly enabled for that Cloud connection.

2. Legacy dashboard default migration is effectively bypassed
   - `DEFAULT_SETTINGS.shareBrowserContext` is `false`, while `loadSettings()` only derives a default when the merged value is not boolean.
   - If defaults are merged before this block, an absent legacy value becomes `false`, so `deriveShareBrowserContextDefault()` never supplies the documented dashboard default of `true`.
   - Preserve whether the property existed in stored settings, or perform migration before applying defaults. Add a full `loadSettings()` migration test rather than testing only the helper.

3. Turn-time overrides bypass the consent gate
   - In `sidepanel.js` around `askHermes()`, `contextOverride?.contextScope` is accepted before `contextScopeForGateway()`:
     `contextOverride?.contextScope || contextScopeForGateway(...)`
   - Any caller supplying an override can retain a non-chat scope after consent is disabled.
   - Select the requested scope first, then always pass it through `contextScopeForGateway()` before context capture/serialization.

## Suggestions

- Update the UI copy: it says both Cloud and dashboard connections default off, while the intended self-hosted dashboard default is on.
- Test capability synthesis independently so `browserContextInline` cannot advertise sharing when the effective scope is Chat-only.

## Checks to run

- Full migration tests using stored settings with no `shareBrowserContext` property.
- Mode-transition tests: dashboard enabled → Cloud must remain off; Cloud enabled → unrelated gateway must not inherit that consent.
- Turn test with `contextOverride.contextScope` while consent is false, asserting no browser-context envelope is sent.
- Compare the complete test suite on base and PR head, and require successful CI or an explicitly accepted baseline-failure comparison before merge.

---

Reviewed signature: `6d6cc7f2aa9e8387e70bccfa52b5913cf4f6873ede6998749e9914192686ddeb`

chtcrack:
Thanks for the PR — the consent-gate direction is right, and defaulting Cloud to OFF is the correct trust posture. That said, I have one blocking concern and a few smaller notes.

**Blocking: `shareBrowserContext` is a single global setting, not per-connection — Cloud can inherit sharing from another connection**

Since the flag is one stored setting rather than a per-connection value, the "Cloud OFF by default" guarantee only holds for a fresh install. The leak path:

1. User connects to a self-hosted remote dashboard → `deriveShareBrowserContextDefault()` returns `true` → the value is stored as `true`.
2. User switches the connection to Hermes Cloud. In `loadSettings`, `typeof settings.shareBrowserContext === 'boolean'` is now `true`, so the derived default is skipped and the stored `true` survives.
3. `saveSettingsFromForm` also preserves the previous value when the new connection isn't context-relevant (`settings.shareBrowserContext === true`), so saving the Cloud config keeps sharing enabled.

Result: page context is shared with a third-party endpoint without the user ever seeing or toggling the Cloud consent. That contradicts the opt-in premise stated in the summary.

Suggested fix: key the consent by connection identity (`connectionMode` + `connectionTransport` + `gatewayUrl`), or reset to `deriveShareBrowserContextDefault(...)` whenever the connection identity changes. Cloud should always start from `false` until explicitly enabled for that connection.

**Non-blocking notes**

- `mergeable_state` is currently `dirty` — the branch needs a rebase on current `main`.
- Test gap: the new tests cover `deriveShareBrowserContextDefault` in isolation, but nothing exercises the `loadSettings` preservation path (the `typeof === 'boolean'` branch) across a connection switch. A regression test simulating a dashboard → Cloud switch and asserting consent resets to `false` would have caught the issue above.
- Minor: the trailing `return true` in `deriveShareBrowserContextDefault` covers API transports where the value is unused by the gate. It's documented, but naming it "connection default" rather than "share default" would make that clearer.

PR #67, opened 2026-08-07
feat: full dashboard companion surface — context upload, events, control, file delivery (Phases 2–7)

abundantbeing:
## Stacked on #66 — read that first

This PR builds **directly on top of [#66 (Phase 0: opt-in browser-context sharing)](https://github.com/abundantbeing/hermes-browser-extension/pull/66)**, which lands the `shareBrowserContext` consent setting and de-gates `contextScopeForGateway()`. Its base is the Phase 0 branch, so **review the diff as the delta on top of #66**; once #66 merges, retarget this one to `main`.

## What this adds (Phases 2–7 of the parity plan)

| Phase | Feature | Extension | Companion plugin | Needs core? |
|---|---|---|---|---|
| **2** | **On-demand context pull** — `browser.context.upload` RPC (debounced out-of-band envelope push) enabling hash-reference turns; **auto-falls back to full inline delivery** if the core rejects the RPC | ✅ | ✅ `put_bcp_v2_payload` + `handle_ws_method` | WS→plugin dispatch (`handle_ws_method` hook) |
| **3** | **Live browser events** — `browser.events.publish` for tab/page lifecycle metadata (never page text), recorded into the owner-scoped event log | ✅ | ✅ | same |
| **4** | **Browser control** — server→client `browser.control` frames: navigate / click / type / scroll / snapshot / extract via `chrome.scripting`; results correlated via `browser.control.result`; **gated by context-sharing consent + `allowBrowserControl` setting + per-action inline approval card** (session auto-approve option); restricted-URL + password-field guards | ✅ | — | outbound frame bridge |
| **6** | **Agent → extension file delivery** — server→client `file.delivery` events render a download card; saved via `chrome.downloads` (4 MiB cap, clipboard fallback) | ✅ | — | outbound frame bridge |
| **7** | **Docs** — README matrix + CHANGELOG | ✅ | | |

## Degradation contract

Every new RPC is **fail-closed**: on cores without WS→plugin dispatch the upload RPC rejects → the extension permanently falls back to full inline delivery (context is never lost), events are best-effort metadata, and control/file delivery simply never fire. This PR is safe to merge before the Hermes-core dispatcher exists.

## Security posture

- Context sharing stays **OFF by default for Hermes Cloud** (third-party endpoint); self-hosted remote dashboards default ON.
- Browser control is triple-gated: sharing consent + explicit `allowBrowserControl` setting + per-action approval in the panel.
- No new permissions: `downloads` and `scripting` were already in the manifest.
- Control refuses restricted schemes/URLs (`isRestrictedUrl`) and refuses to type into password fields.
- Events carry metadata only; uploads stay within the BCP 48k budget; file delivery caps at 4 MiB.

## Test results

`node --test tests/*.test.mjs` → **755 pass / 9 fail** — the 9 failures are pre-existing on `main` in this environment (verified identical set on a clean checkout): firefox-build, i18n-runtime, inline-draft-policy, site-adapters, context-menu-editor*, content-extraction-core, assistant-clipboard.

## Follow-ups (outside this repo)

- **Hermes core**: WS dispatcher → plugin `handle_ws_method` routing + outbound control/file frame bridge (spec'd in the plan doc).
- Companion-plugin control tools (`browser_navigate` etc.) land once the core bridge exists.
- Manual E2E: dev-load against a real Cloud agent, flip the toggle, verify `delivery: full` arrives agent-side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added opt-in **browser-context sharing for Hermes Cloud and no-key Remote (dashboard WebSocket) connections** (`shareBrowserContext`): Cloud defaults OFF (third-party endpoint; sharing is a per-connection consent decision), self-hosted remote dashboards default ON. When enabled, Follow active tab, Pin current tab, and per-tab prompt selection work over the dashboard socket with full BCP v2 delivery, and the companion plugin's context tools (`browser_context_status`, `browser_get_context`) activate.
+- Added the dashboard WS companion surface: `browser.context.upload` (out-of-band context push enabling hash-reference turns), `browser.events.publish` (tab/page lifecycle metadata), `file.delivery` (agent → extension downloadable file cards via `chrome.downloads`), and `browser.control` (opt-in agent-driven browser actions with per-action approval cards). All degrade to full-inline delivery on cores without the RPCs.
+- Added `allowBrowserControl` opt-in setting: lets the agent navigate/click/type/scroll/snapshot/extract in the active tab, gated by context-sharing consent, an explicit setting, and per-action approval.
+
+### Changed
+
+- `contextScopeForGateway()` only forces Chat-only for dashboard gateways when context sharing is not consented; all scope call sites (startup, `applyContextScope`, turn-time) honor the setting.
+- Remote-dashboard capability synthesis reports `browserContextInline` and `browserControl` honestly; `app.js` context labels show "Tab context shared" vs "Chat only".
+
 ## [0.2.0] - 2026-07-21
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Browser-native side panel for [Hermes Agent](https://hermes-agent.nousresearch.c
 
 ## What it is
 
-Hermes Browser Extension is not a browser chatbot. It is a Chrome/Edge/Chromium side panel for the real Hermes Agent runtime. Choose a local gateway, attach to a signed-in Hermes Cloud agent tab, or connect to a self-hosted remote API/dashboard. Local and remote API connections can use the models, tools, skills, sessions, memory, and MCP servers already configured in Hermes; Cloud and dashboard-ticket connections are intentionally Chat-only.
+Hermes Browser Extension is not a browser chatbot. It is a Chrome/Edge/Chromium side panel for the real Hermes Agent runtime. Choose a local gateway, attach to a signed-in Hermes Cloud agent tab, or connect to a self-hosted remote API/dashboard. Local and remote API connections can use the models, tools, skills, sessions, memory, and MCP servers already configured in Hermes; Cloud and dashboard-ticket connections are Chat-only until you opt in to context sharing in Settings.
 
 This repo is specifically for the **Hermes Browser Extension**: the Chrome/Edge/Chromium side-panel integration for Hermes Agent.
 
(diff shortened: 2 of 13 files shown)


abundantbeing:
## Hermes Agent Review

Summary

PR #67 adds a high-risk dashboard companion surface covering context upload, browser events, file delivery, and browser control. The live PR is mergeable into its stacked base, but GitHub shows no status checks. Merge should be blocked until the protocol and control-security issues below are fixed.

Blocking issues

1. Context upload sends the wrong object and cannot work
   extension/sidepanel.js:8493 passes the full hermes.browser.turn.v2 envelope to pushBrowserContextUpload(), but companion-plugin/context_store.py:260 accepts only a hermes.browser.context.v1 payload. Every upload handled by this plugin is therefore rejected.

2. Upload failures and timing can cause context loss
   extension/sidepanel.js:8120 only falls back when client.request() rejects. A normal RPC result such as {ok: false, error: "storage-unavailable"} is ignored. Upload is also delayed by 800 ms and not awaited before subsequent turns may switch to reference delivery. The extension can consequently send a reference before any context was stored, contradicting the “context is never lost” degradation contract.

3. Uploaded context cannot be reused across turns
   companion-plugin/context_store.py:52 requires session_id, turn_id, and task_id, then status/consume operations require the exact same ContextOwner. An out-of-band upload intended for later turns will either be rejected as unowned or become inaccessible when the next turn has different turn/task IDs. The upload needs an authenticated session-scoped ownership model or an explicit, secure handoff to the consuming turn.

4. Declining can accidentally enable automatic approval
   extension/sidepanel.js:8256 sets browserControlAutoApprove whenever the checkbox is checked, including when the user clicks Decline. That makes later control requests execute without approval. Set auto-approve only after an accepted request.

5. “Approve for this session” survives session and connection changes
   browserControlAutoApprove is not reset when the WebSocket closes, the gateway changes, or a different stored session is selected. Reset it at every connection/session boundary and when browser control or context sharing is disabled.

6. Restricted-page protection is incomplete
   extension/sidepanel.js:8323 applies isRestrictedUrl() only to navigation destinations. snapshot, extract, click, and type operate on the current active tab without checking its URL. An already-open banking, billing, password-manager, medical, or other restricted HTTP(S) page can therefore be read or controlled, especially dangerously after auto-approval.

7. Control and file events are not bound to the active session
   Typed gateway events include sessionId, but handleBrowserControlEvent() and handleFileDeliveryEvent() ignore it. Requests from another session on the same socket could render files or control the active tab. Require the event session to match the connection’s live or stored session identity before processing it.

8. The 4 MiB file limit is applied too late
   extension/sidepanel.js:8140 fully decodes and allocates the unbounded base64 input before checking bytes.byteLength. A large gateway event can exhaust side-panel memory. Reject oversized encoded input before atob(), then retain the decoded-size check.

Suggestions

1. Validate uploaded payloads server-side using the same strict BCP schema path as normal turn envelopes; do not trust the extension’s “pre-validation.”

2. Preserve and verify the separately supplied context_hash. The plugin currently ignores params.context_hash and reads payload.context_hash, which the context payload does not contain.

3. Show the text being typed in the approval card so the user can review the complete action.

4. Propagate nested control failures correctly. click/type/extract can return {ok:false}, but handleBrowserControlEvent() currently reports the outer result as ok:true.

5. Sanitize delivered filenames to a safe basename and revoke generated object URLs after download.

Checks to run

1. Add an integration test using the exact object sent by pushBrowserContextUpload(), not a hand-built context payload.

2. Test successful upload, {ok:false} RPC responses, unknown-method rejection, delayed upload, and a second same-hash turn arriving before upload completion.

3. Test upload ownership and retrieval across two different turn/task IDs in the same authenticated session.

4. Test decline with “Approve for this session” checked, reconnect/session switching, and disabling then re-enabling context sharing.

5. Test every control action against an already-open restricted URL and against an event carrying a different sessionId.

6. Test oversized base64 input before decoding, malformed payloads, unsafe filenames, and repeated downloads for object-URL cleanup.

7. Run the full test suite and required browser E2E checks. No GitHub checks are present in the event data, so this review does not approve the PR or treat the author-reported results as verified.

---

Reviewed signature: `68be63707a37d57323ae9e2d094f6f78c8b54816e020d8e5ad20ab75a723aad5`

Issue #73, opened 2026-08-16
Chatbox: /commands menu hides long prompt text and forces scrolling


kidclone3:
## Bug description
When the chatbox `/commands` UI is visible while composing a long prompt, it covers or hides part of the entered text. Users must scroll to keep reading what they are typing.

## Steps to reproduce
1. Open the chatbox.
2. Trigger the `/commands` UI.
3. Type or paste a prompt long enough to span several lines.
4. Observe that the commands UI obscures part of the prompt and requires scrolling.

## Expected behavior
The full composed prompt should remain visible while `/commands` is open, without unnecessary scrolling.

## Actual behavior
The `/commands` UI hides part of a long prompt, forcing the user to scroll while typing.

abundantbeing:
## Hermes Agent Issue Triage

## Summary

Composer layout bug: the `/commands` menu overlaps or reduces the visible area for multiline prompt text, forcing users to scroll while typing.

## Triage

- Category: UI/layout bug
- Component: Side panel composer and commands menu
- Priority: P3, low severity
- Risk: Low security and data risk; moderate usability/accessibility impact for long prompts
- Current labels are mostly appropriate: `type/bug`, `comp/sidepanel`, `p3`
- `type/support` appears unnecessary unless maintainers need environment-specific diagnosis

## Questions / missing info

- Extension version or commit
- Browser and browser version
- Side-panel width, window size, and display scaling
- Approximate prompt length or line count that triggers the issue
- Whether the menu overlays the textarea or only reduces its usable height
- Whether this occurs at all widths and zoom levels
- Screenshot or short recording showing the overlap and scroll position

## Suggested next action

Reproduce across narrow and standard side-panel widths, browser zoom levels, and pasted versus typed multiline text. Inspect the composer/menu positioning, textarea auto-growth, overflow, and reserved menu space. Define acceptance criteria that the caret and active prompt line remain visible while `/commands` is open, without the menu covering composed text.

---
_Automated review by Hermes Agent. Diffs/issues are treated as untrusted input._

Reviewed signature: `8299fad40c7c269acb2342dba2d0efd592c6902ff672a22cb2b049bdc90fd5cb`

abundantbeing:
[...]

Issue #69, opened 2026-08-08
Desktop app and Chrome extension desync when sharing a session — after Stop, the agent keeps running invisibly in a stale copy of the conversation ("ghost loop")


Prediction6767:
## The bug in plain words

This is a bug about the desktop app and the Chrome extension side panel **desyncing** when they share the same session. When they desync, there are effectively **two copies of the conversation**: the copy you see (which shows the agent stopping when you press Stop) and a stale copy the agent is actually running on (which never received your Stop). The agent keeps working invisibly — moving the mouse, typing — while every surface shows it as idle or stopped.

## The events (what happened)

1. The user attached an existing desktop session (500+ messages, ~10 hours of context) to the side panel. The picker asked "keep this session as desktop" vs "open a new browser session" — no explanation of what either meant. The user chose **keep as desktop**, assuming they could keep talking with the conversation after the browser job.
2. The user pinned a GitHub "New Issue" tab and asked the agent to fill the form. The agent drove the real browser via OS-level input automation (PowerShell SendInput — real mouse/keyboard).
3. Mid-job, the user clicked **Stop** in the side panel and typed "bro stop" / "bro abort the mission". The panel showed the agent reply "Stopped. No more inputs."
4. The mouse and keyboard **kept acting on their own** for ~4 more minutes, despite both Hermes UIs (desktop app and browser extension) showed idle, and the chat log appeared reverted a couple of turns on its own.
5. The loop only ended when the user cut the network to the model provider.

## The diagnosis (evidence)

1. **The Stop button is client-side only.** `dist/app.js`: `els.stopRun.addEventListener('click', () => activeAbortController?.abort())`. It aborts the panel's own stream and sends **nothing** to the server — the server-side turn keeps running with no UI.
2. **The server ran concurrent turns on the same session.** Gateway log: `Skipping Relay instrumentation for concurrent Hermes turn <session>:<turn_a> ... and <session>:<turn_b>` — plus `Stored system prompt for session <id> is null; rebuilding from scratch`.
3. **Two divergent contexts alternated.** Model-call logs show a call containing "bro stop" producing the "Stopped" reply, and the next call (same session, seconds later) with a context **without** the stop message resuming the task ("Version filled ✓ — scroll and retry"). The stop command existed in only one copy of the conversation.
4. **Session state was rewritten mid-incident.** Two message rows were deleted from the session DB, and both UIs showed the log reverted — consistent with two writers on one session.

## How to reproduce

[...]

PR #76, opened 2026-08-17
fix: scroll prompt caret into view when commands menu opens; warn on large shared sessions

iamformlesslol:
## What this fixes\n\nTwo open issues:\n\n### #73 — /commands menu hides long prompt text\n\nThe quick-more-menu floats absolutely over the textarea. When a prompt spans several lines, the menu covers the caret and the user has to scroll to see what they are typing.\n\nFix: scrollInputToCaret() now runs when the menu opens — it sets the caret position and scrolls the textarea so the active line stays visible.\n\n### #69 — Ghost loop on shared sessions\n\nThe session ownership notice warned generically about two surfaces writing one session. For large sessions (100+ messages), it now explicitly calls out the ghost-loop/desync risk and steers users toward a new chat.\n\n## Changes\n\n- extension/sidepanel.js — setQuickCommandMenuOpen() calls scrollInputToCaret() on open; new scrollInputToCaret() helper\n- extension/lib/session-ownership.mjs — large-session detail text warns about ghost loops\n\n## Verification\n\n- npm run check:js — pass\n- npm run build — pass (self-contained check passed, dist built)\n- npx eslint — 0 errors, 60 warnings (all pre-existing)

### DIFF
--- a/extension/lib/session-ownership.mjs
+++ b/extension/lib/session-ownership.mjs
@@ -80,11 +80,15 @@ export function sessionOwnershipNotice({ session = {}, expectedSource = '' } = {
   const newChatLabel = normalized(expectedSource) === SESSION_SURFACE_SOURCES.FULL_TAB
     ? 'New Hermes Web chat'
     : 'New Browser chat';
+  const largeSession = count !== null && count > 100;
+  const detail = largeSession
+    ? `${countDetail} Large shared sessions can desync: Stop in one surface may not reach the other, causing the agent to keep running invisibly. Start a new chat to avoid ghost loops.`
+    : `${countDetail} Two Browser surfaces writing the same session can overwrite or reorder transcript updates. Start a new chat, or continue here intentionally.`;
   return Object.freeze({
     sourceLabel,
     messageCount: count,
     title: `${sourceLabel} session selected`,
-    detail: `${countDetail} Two Browser surfaces writing the same session can overwrite or reorder transcript updates. Start a new chat, or continue here intentionally.`,
+    detail,
     newChatLabel,
     continueLabel: `Continue in ${sourceLabel}`,
   });
--- a/extension/sidepanel.js
+++ b/extension/sidepanel.js
@@ -5657,6 +5657,17 @@ function setQuickCommandMenuOpen(open) {
   els.quickMoreMenu.hidden = !open;
   els.commandMenuButton?.setAttribute('aria-expanded', String(Boolean(open)));
   if (!open) clearQuickCommandDetail();
+  else scrollInputToCaret();
+}
+
+function scrollInputToCaret() {
+  const input = els.input;
+  if (!input || !input.setSelectionRange) return;
+  requestAnimationFrame(() => {
+    const pos = input.selectionStart || input.value.length;
+    input.setSelectionRange(pos, pos);
+    input.scrollTop = input.scrollHeight;
+  });
 }
 
 function clearQuickCommandDetail() {


abundantbeing:
Automated agent review from the Hermes Browser side panel (Hermes Agent, model deepseek/deepseek-v4-flash-0731 via Nous). Reviewed at PR head b9483e9 against live main 1a34e19. The diff applies cleanly with no conflicts, both changed files pass node --check, and the existing session-ownership tests stay green (6/6) against the PR's module.

## Summary

Small, clean two-part change. The ownership-notice copy is safe to merge as-is. The caret-scroll helper has two correctness problems worth fixing before merge: a falsy-zero caret bug, and a scroll target that does not actually match the stated goal.

## Important

1. Caret jumps to end when selectionStart is 0 (extension/sidepanel.js, scrollInputToCaret)
`const pos = input.selectionStart || input.value.length;`
When the caret is at the very start of the prompt, selectionStart is 0, which is falsy, so pos becomes input.value.length and opening the /commands menu jumps the caret to the END of the text. Use a nullish check instead: `input.selectionStart ?? input.value.length` (or a typeof check).

2. Scroll target is the content bottom, not the caret line
`input.scrollTop = input.scrollHeight` pins the viewport to the bottom of the textarea content. For a caret mid-prompt it does not bring the active line into view. And the menu floats over the composer's bottom region (position absolute, bottom 42px, up to 386px tall), so in the common typing case (caret on the last line) this scrolls the caret line INTO the covered zone rather than above it. The robust fix measures the caret's line position (mirror-div technique or caret rect) and sets scrollTop so the active line clears the menu's overlap; at minimum, verify in a real render that the last line is visible above the menu.

## Suggestions

3. setSelectionRange(pos, pos) also collapses any text selection the user had in the prompt when the menu opens. Consider preserving selectionStart/selectionEnd when a selection exists.

4. Ghost-loop copy vs shipped behavior (extension/lib/session-ownership.mjs)
The new text claims "Stop in one surface may not reach the other, causing the agent to keep running invisibly." Current main ships transport-aware server-side Stop (stopCurrentTurn / stopActiveRun with runtime-acknowledgement handling and queued-turn holds), so the #69 client-side-only-Stop mechanism is largely addressed. The residual risk on a shared session is two surfaces writing one session, which the existing copy already covers. Suggest rewording around concurrent-writer desync rather than Stop delivery, and keep #69 open: this change warns users away from large shared sessions, it does not fix the underlying writer race.

5. No test covers the new branch. tests/session-ownership.test.mjs covers counts 2 and unknown, but not count > 100. Boundary nit: the code uses > 100 (so exactly 100 messages gets the generic copy) while the PR body says "100+". Add tests for 100 and 101.

## Verified

- Base 1a34e19 equals live main, mergeable state clean, diff applies via git apply --check
- node --check passes for both changed files at PR head
- tests/session-ownership.test.mjs: 6/6 pass against the PR's module
- Did not run the repo's full lint/build, so the "0 errors, 60 warnings (all pre-existing)" claim is reported but not independently confirmed

No security concerns in this diff. Thanks for the contribution.

abundantbeing:
Closing this PR for now due to substantial base branch drift and implementation issues.

### 1. Base Branch Drift
 has since merged **Phase 6 (browser and tab control + gateway pairing)** (), introducing major architectural changes across the service worker, leases, and the full multi-language locale catalog. Because this PR was branched from pre-Phase-6 , merging or fast-forwarding would clobber Phase 6 control infrastructure.

### 2. Implementation Blockers
- **Falsy-Zero Caret Bug:**  evaluates  as falsy, jumping the caret to the end of the text if the cursor is at the beginning of the prompt. A nullish coalescing check () or explicit  check is required.
- **Scroll Target Bounds:** Setting  pins the viewport to the absolute bottom of the textarea rather than aligning to the caret line. Because the  menu floats above the composer, this can push the caret line directly under the menu overlay.
- **Session Ownership Copy Drift:** The ownership notice mentions client-only stop delivery; however, Phase 5/6 landed transport-aware stopping ( /  with runtime ACKs) and single-writer leases, so the warning text is out of sync with current architecture.

Feel free to open a fresh, targeted PR rebased on latest  with the caret calculation fixes if you'd like to revisit the command menu caret behavior!

abundantbeing:
Thank you @iamformlesslol for catching this issue and opening the PR!

The underlying issue—keeping the active editing caret line visible in the textarea when the floating  menu opens—has been resolved and merged directly into  in commit [](https://github.com/abundantbeing/hermes-browser-extension/commit/3ea6c90).

### What Landed ():
- **Caret Pinning Fix:** Replaced  with an explicit non-zero check , preventing cursor jumps to the end of prompt when the caret is at index .
- **Proportional Viewport Alignment:** Instead of forcing  (which could push the caret behind the floating menu),  now calculates proportional line-height scroll offsets to ensure the active typing line remains visible above the overlay.
- **Selection Preservation:** Retains active  /  text highlights without collapsing them.

Closing this PR as resolved via . Thanks again for the contribution!

abundantbeing:
imagehbe

PR #90, opened 2026-08-28
fix(sidepanel): prevent Edge panel reloads during tab changes

forgegod:
## What does this PR do?

Fixes Edge side-panel reloads caused by Manifest V3 worker lifecycle races and native side-panel fallback behavior.

- Wait for persisted panel residency before applying options after tab activation.
- Reconfigure panel residency only when the cached mode actually changes, including when Edge omits `oldValue` from a storage event.
- Preserve Edge's browser-owned native side-panel action if a manual open rejects instead of opening a full extension tab.
- Route successful manual connection tests through canonical startup readiness so the startup gate dismisses only after model and session readiness complete.
- Settle automatic-pairing test feedback from the pairing generation.

## Related Issue / Phase

No matching upstream issue or PR exists. This is a browser-compatibility fix reproduced against Edge side-panel behavior.

## Type of Change

- [x] 🐛 Bug fix
- [x] 🌐 Browser compatibility
- [x] ✅ Tests

## Changes Made

- `extension/background.js`
  - Prevents activation-time residency application before storage hydration.
  - Preserves global panels across non-residency settings writes.
  - Retains the native browser side-panel action when manual opening rejects.

- `extension/sidepanel.js`
  - Uses canonical readiness after a successful manual API connection test.
  - Finishes automatic-pairing test feedback from the active pairing generation.

- Tests
  - Cover global activation during delayed residency hydration.
  - Cover Edge storage changes without `oldValue`, API-key saves, and native-action failures.
  - Cover startup readiness after successful manual connection tests.

## How to Test

1. `PYTHON="$(uv python find)" pnpm run verify`
2. `pnpm run lint`
3. `pnpm run build`
4. Reload the unpacked extension from `dist/` in Edge.
5. Enable **Keep open across tabs**, then:
   - save a changed API key and run **Test connection**;
   - navigate to a localhost dashboard route and return to another tab;
   - open and leave a redirect-heavy OIDC login tab;
   - click the extension toolbar icon.
6. Confirm that the native side panel remains intact, startup proceeds after a successful test, and no full extension tab opens.

## Browser Evidence

- Microsoft Edge 151.0.4129.107 (64-bit)
- Firefox 154.0.1 (64-bit), chat/context/sidebar validation

Firefox remains chat-and-context only; real-tab control is Chromium-only.

## Verification

- `pnpm run lint` — passed
- `PYTHON="$(uv python find)" pnpm run verify` — 1282/1282 passed
- `pnpm run build` — passed
- Firefox package lint — 0 errors, 16 existing `innerHTML` warnings
- Live Edge CDP check: opened the LiteLLM localhost route, returned to the prior debug tab, and observed no side-panel frame navigation, execution-context destruction, or startup reset.

## Screenshots / Logs

A public-safe CDP recording was captured for review. It shows the global panel ready, a local-dashboard navigation/return event with zero panel reload signals, and the panel still ready afterward.

edge-panel-residency-cdp

## Checklist

- [x] I've read the Contributing Guide
- [x] My commit messages follow Conventional Commits
- [x] All tests pass and project verification completed
- [x] JavaScript syntax, manifest, and locale validation pass
- [x] Context consent and redaction rules are preserved
- [x] No hardcoded secrets, tokens, or unredacted credentials
- [x] Tested target browser(s)

### DIFF
--- a/extension/background.js
+++ b/extension/background.js
@@ -85,6 +85,7 @@ const browserControlRuntime = createBrowserControlRuntime({
   },
 });
 let cachedPanelResidencyMode = DEFAULT_PANEL_RESIDENCY_MODE;
+let panelResidencyHydrated = false;
 
 const controllerConnector = createControllerConnector({
   fetchImpl: globalThis.fetch?.bind(globalThis),
@@ -464,6 +465,8 @@ async function refreshPanelResidencyModeFromStorage() {
   } catch (error) {
     console.warn('[Hermes Browser] Could not read panel residency setting:', error);
     cachedPanelResidencyMode = DEFAULT_PANEL_RESIDENCY_MODE;
+  } finally {
+    panelResidencyHydrated = true;
   }
   return cachedPanelResidencyMode;
 }
@@ -534,7 +537,16 @@ async function configureSidePanel() {
 }
 
 function reapplyPanelResidencyForTab(tabId) {
-  applyPanelResidencyMode(cachedPanelResidencyMode, { tabId })
+  // The worker starts with the tab-attached default. Do not overwrite an
+  // existing global panel while its persisted residency mode is still loading.
+  if (!panelResidencyHydrated) {
+    return refreshPanelResidencyModeFromStorage()
+      .then(() => reapplyPanelResidencyForTab(tabId));
+  }
+  // A global panel has one default path for the window. Reapplying identical
+  // options for every activated tab can recreate that document in Edge.
+  if (cachedPanelResidencyMode === PANEL_RESIDENCY_MODES.GLOBAL) return;
+  return applyPanelResidencyMode(cachedPanelResidencyMode, { tabId })
     .catch((error) => console.warn('[Hermes Browser] Could not apply panel residency setting:', error));
 }
 
@@ -728,9 +740,13 @@ function openHermesPanelFromGesture(tab) {
         }
         return true;
       })
-      .catch(() => openHermesPanelFallback(tab, browserId, panelUrl));
-  } catch {
-    return openHermesPanelFallback(tab, browserId, panelUrl);
+      .catch((error) => {
+        console.warn('[Hermes Browser] Native side panel open rejected; preserving the browser-owned side-panel action instead of opening a full tab:', error);
+        return true;
+      });
+  } catch (error) {
+    console.warn('[Hermes Browser] Native side panel open threw; preserving the browser-owned side-panel action instead of opening a full tab:', error);
+    return true;
   }
 }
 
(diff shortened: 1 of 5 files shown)


abundantbeing:
## Hermes Agent Review

Consolidating now. I've read the full context of all three changed surfaces. Remaining verification before the verdict: (1) can any settings writer omit `panelResidencyMode` (the new onChanged handler normalizes a missing key to tab-attached, which would close a live global panel), and (2) do the touched test files actually pass at PR head in my isolated worktree.

Review complete. Verification done at PR head 4f26eb2 in an isolated worktree (Jon's primary worktree untouched); both touched test suites pass locally there (43/43), but that's reviewer-local evidence only — the event data contains no CI/check results, so no approval.

## Summary

Solid, well-targeted compatibility fix with real harness-driven regression tests. The four behavior changes are coherent: (1) activation-time reapply is now gated on storage hydration and no-ops for global residency (correct — Chromium no-ops identical `setOptions` anyway, Edge recreates the document); (2) `onChanged` now recomputes `changed` from the cached mode instead of trusting `oldValue`, which fixes Edge omitting `oldValue`; (3) Chromium gesture failures no longer open a full extension tab; (4) manual connection tests now route through `runPanelConnectionReadiness` instead of a bespoke `/v1/models` probe. I traced the risky paths: the pairing branch returns early after `connectApiWithPairing` (which itself runs readiness at sidepanel.js:9587 and transitions READY), so no double-flash or stale-busy; `bindSession` inside readiness covers the removed `ensureHermesSession` call; no security surface changes (no new permissions, no network, no untrusted-input paths touched).

## Blocking issues

None. Nothing blocks merge.

## Suggestions

1. `extension/background.js:917-924` — the new `onChanged` handler recomputes the cached mode on *every* `hermesBrowserSettings` write, normalizing a missing `panelResidencyMode` to tab-attached. The old code ignored key-less writes. Any writer that persists a settings object without the key (legacy shape, future writer, or a write racing hydration) flips global → tab-attached and closes the user's live global panel. Current writers I checked spread stored settings and `DEFAULT_SETTINGS` includes the key, so this is latent, not live — but the cheap hardening is: only recompute when `changes.hermesBrowserSettings?.newValue?.panelResidencyMode != null` (the Edge no-`oldValue` fix still works because its `newValue` contains the key), and add a regression test for a key-less write while cached mode is global.
2. The removed `/v1/models` degraded-probe path means a health-OK gateway with broken model auth no longer shows the "connected with runtime warning" / `DEGRADED` state from a manual test — it shows READY with fallback model metadata. Likely intentional consolidation onto canonical readiness; confirm it's an accepted UX change, not an accident of the rewrite.
3. `extension/background.js:743-749` — if `sidePanel.open()` rejects persistently on some Chromium fork, the toolbar icon becomes a silent no-op (no panel, no tab). Accepted trade-off per the PR body, but consider a one-time `console.warn` throttle or a settings-dialog hint if telemetry shows rejections.

## Checks to run

The event data shows no GitHub checks; do not treat PR-body claims as results.

- `npm run verify` (tests + `check:js` + `check:manifest`)
- `npm run lint`
- `npm run build`
- Reload `dist/` unpacked in Edge with global residency enabled: toggle an API-key save, switch tabs across a localhost route and a redirect-heavy OIDC tab, and confirm no panel reload, no full extension tab on toolbar click
- Reviewer-local (head 4f26eb2, isolated worktree, primary tree untouched): `node --test tests/browser-runtime.test.mjs tests/browser-control-turn-integration.test.mjs` → 43/43 pass

---

Reviewed signature: `8e5b10496b058f638f06e219607319f7b8c1f7117ca481137dbf447302405731`

PR #101, opened 2026-09-06
fix(sidepanel): keep the model picker inside the viewport

qinxianhahaha:
## Problem

Opening the model picker in the side panel can render it with its top rows above the top of the viewport. The **search box and the provider selector are cut off entirely** — the menu appears to open straight onto the model list, with only the provider row's horizontal scrollbar peeking in at the top edge. There is no way to recover them: `.model-menu` is `overflow: hidden`, so the clipped rows cannot be scrolled back.

That makes it impossible to switch providers or search models from the side panel; the user is stuck with whatever provider happens to be selected.

## Cause

The menu is bottom-anchored above the composer dock:

```css
.model-menu, .context-popover {
  position: absolute;
  bottom: calc(var(--hermes-bottom-dock-height, 0px) + 6px);
}
.model-menu {
  max-height: min(480px, calc(100vh - 176px));
  overflow: hidden;
}
```

The `bottom` offset tracks the live dock height (`--hermes-bottom-dock-height`, kept up to date by `updateDockFloatingAnchor()`), but `max-height` reserves a hardcoded **176px**. The real dock is routinely much taller than that once the attachment row, the `Enter sends…` help text, the composer actions and the desktop bar are laid out — around 300px in a normal side panel.

So the box is placed at `bottom + 480px`, which lands above the viewport top, and `overflow: hidden` makes the overflowing header unreachable.

Measured in a side panel at 635x760 with a 304px dock, before the change:

```
dock=304px vh=673  menu.top=-117  menu.height=480
searchTop=-116  providerListTop=-81
```

The 58px provider row sits entirely off-screen (`-81 … -23`).

A secondary contributor: the grid rows `minmax(58px, auto) minmax(96px, 1fr) auto auto` give the model list a 96px floor and the runtime-options block an unbounded `auto` row, so when space is tight those rows push the search box and provider selector out of the box instead of shrinking.

## Fix

- Cap `.model-menu` / `.context-popover` against the live dock height rather than a fixed guess:
  `max-height: min(480px, calc(100vh - var(--hermes-bottom-dock-height, 0px) - 18px))`.
  The `0px` fallback keeps the pre-measurement first paint correct.
- Let the model list and the runtime-options rows shrink and scroll (`minmax(0, 1fr)` / `minmax(0, auto)`, plus `min-height: 0; overflow-y: auto` on `.model-options-list`) instead of pushing the header off-screen. Applied to both the chat and the Hermes Assist grid templates.
- Give `.context-popover` the same max-height and `overflow-y: auto`; it shares the same bottom anchor and the same failure mode.

CSS-only, no markup or behavior change.

After, same layout:

```
dock=304px vh=673  menu.top=12  menu.height=351
searchTop=13  providerListTop=48
```

## Verification

Rendered the real `sidepanel.html` + `sidepanel.css` with a populated model menu (4 providers, 13 models, 7 effort levels) in headless Chrome.

| viewport | dock | `menu.top` before | `menu.top` after |
| --- | --- | --- | --- |
| 360x700 | 304px | off-screen | 12px |
| 400x520 | 304px | off-screen | 12px |
| 635x760 | 304px | -117px | 12px |
| 635x1400 | 304px | ok | 523px |

Search box and provider selector are fully visible in every case, for both the chat and the Hermes Assist (`data-selection-target="assist"`) targets.

- `npm run check:js` — pass
- `npm run check:manifest` — pass
- `npm run lint` — 0 errors (873 pre-existing warnings)
- `npm test` — 1197/1207 pass. The 10 failures are in `tests/companion-plugin.test.mjs` and related Python-companion suites and reproduce identically on unmodified `main` in this environment; they are unrelated to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed the side-panel model picker opening partly above the viewport, which hid its search box and provider selector with no way to scroll them back into view. The bottom-anchored popovers now reserve the live composer dock height instead of a fixed 176px, and the model list and runtime options shrink and scroll instead of pushing the header off-screen.
+
 ## [0.2.0] - 2026-07-21
 
 ### Added
--- a/extension/sidepanel.css
+++ b/extension/sidepanel.css
@@ -3072,12 +3072,19 @@ textarea:focus, input:focus, select:focus { border-color: var(--hermes-accent); 
 .context-popover {
   position: absolute;
   bottom: calc(var(--hermes-bottom-dock-height, 0px) + 6px);
+  /* Keep bottom-anchored popovers inside the viewport: the space above the dock
+     is what is actually available, so reserve the live dock height instead of a
+     fixed guess. Otherwise a tall composer pushes the popover's top rows (search
+     box and provider picker) above the viewport, where `overflow: hidden` makes
+     them unreachable. */
+  max-height: calc(100vh - var(--hermes-bottom-dock-height, 0px) - 18px);
   z-index: 70;
 }
+.context-popover { overflow-y: auto; }
 .model-menu {
   display: grid;
-  grid-template-rows: auto minmax(58px, auto) minmax(96px, 1fr) auto auto;
-  max-height: min(480px, calc(100vh - 176px));
+  grid-template-rows: auto minmax(58px, auto) minmax(0, 1fr) minmax(0, auto) auto;
+  max-height: min(480px, calc(100vh - var(--hermes-bottom-dock-height, 0px) - 18px));
   overflow: hidden;
 }
 .model-menu-head { display: none; }
@@ -3086,7 +3093,7 @@ textarea:focus, input:focus, select:focus { border-color: var(--hermes-accent); 
   inset: auto 12px auto 12px;
   width: auto;
   max-height: min(560px, calc(100vh - 96px));
-  grid-template-rows: auto auto minmax(58px, auto) minmax(120px, 1fr) auto auto;
+  grid-template-rows: auto auto minmax(58px, auto) minmax(0, 1fr) minmax(0, auto) auto;
   z-index: 100;
   box-shadow: 10px 10px 0 rgba(var(--hermes-shadow-rgb),0.2);
 }
@@ -3276,6 +3283,8 @@ textarea:focus, input:focus, select:focus { border-color: var(--hermes-accent); 
 .model-option-meta,
 .session-option-meta { color: rgba(var(--hermes-fg-rgb),0.58); font-size: 10px; }
 .model-options-list {
+  min-height: 0;
+  overflow-y: auto;
   border-top: 1px solid rgba(var(--hermes-fg-rgb),0.18);
   padding: 7px 10px 8px;
   background: rgba(0,0,0,0.10);


abundantbeing:
## Hermes Agent Review

### Summary

CSS-only fix for a real side-panel bug. `.model-menu` is bottom-anchored with `--hermes-bottom-dock-height`, but `max-height` still subtracted a hardcoded `176px`. A ~300px composer dock then pushed the search box and provider row above the viewport, and `overflow: hidden` made those rows unrecoverable.

Pairing `max-height` to the live dock variable, and letting the model-list / runtime-options rows shrink (`minmax(0, 1fr)` / `minmax(0, auto)` plus `min-height: 0; overflow-y: auto` on `.model-options-list`), is the right approach. The same dock-aware cap plus `overflow-y: auto` on `.context-popover` matches the shared bottom anchor.

Current `main` still has `max-height: min(480px, calc(100vh - 176px))` and `minmax(96px, 1fr)`, so the bug is still present upstream. No JS, permission, or network changes. Security risk is negligible.

This review does not approve. GitHub reports the PR as not mergeable, and there are no check runs on the head SHA.

### Blocking issues

1. Rebase onto current `main` before merge. GitHub reports `mergeable: false` / `mergeable_state: dirty`, and the branch is 1 commit ahead / 35 behind `main`. `CHANGELOG.md` `[Unreleased]` on `main` already has several Fixed bullets, and `extension/sidepanel.css` has grown since this branch was cut. The intended selectors still exist on `main`, so the fix should re-apply, but the current patch cannot land as-is.

No other code-level merge blockers in the proposed CSS once it is retargeted.

### Suggestions

- After rebase, add the changelog bullet under the existing `[Unreleased] ### Fixed` list. Do not replace that section.
- Hermes Assist still uses `max-height: min(560px, calc(100vh - 96px))`. Assist placement also writes inline `max-height` / `top` / `bottom` in JS (`positionAssistModelMenu()`), then clears those properties when returning to chat. The assist grid shrink is still useful; confirm chat mode picks up the new dock-aware CSS after that inline style is removed.
- `.model-provider-list` still has `min-height: 58px`. On an extremely short panel, search + provider + footer could still exceed the capped height while `.model-menu` stays `overflow: hidden`. Optional fallback: allow the menu itself to scroll if min-content overflows.
- The extra `.context-popover { overflow-y: auto; }` rule can be folded into the preceding block.

### Checks to run

- Rebase onto `main` and confirm both hunks still apply to the current `.model-menu`, `.context-popover`, and `.model-options-list` rules.
- Side panel visual: about 360x700 and 635x760 with a tall composer (attachments, help text, desktop bar). Search and provider row must be fully visible; the model list should scroll.
- Open the context-window popover in the same layout and confirm it stays in view and scrolls.
- Settings, Hermes Assist model picker: search/provider still reachable.
- Firefox side panel (`100vh` in a sidebar document).
- `npm run check:js` and `npm run lint` after rebase. This review has no check-run data and does not treat the PR body's test counts as evidence. The change is CSS-only.

---

Reviewed signature: `ed4c0b29ba1f8a4252fa2fe30b59d306e7ca7872de152c82eaa6825707df49b5`

abundantbeing:
## Hermes Agent Review

I'll load the review and GitHub workflow skills first, then pull PR #101 from the live repo so the comment is grounded in the actual diff.

Some skill payloads were truncated. I’ll pull the review-watch workflow and the live PR #101 state next.

Next I’ll pull Naminé’s review taxonomy and the live PR #101 metadata, diff, and files.

I have the comment/label contract. Next I’ll inspect the live PR, surrounding CSS, and how the dock-height variable is actually set.

Local tree is still main. I’ll pull the PR head files, comments, checks, and the dock-height wiring next.

The PR is 35 commits behind and marked dirty. I’ll compare current main’s model-menu CSS and the dock-height logic before deciding what actually blocks.

The CSS approach looks right, but GitHub still marks the PR conflicting. I’ll post the review comment and then confirm it landed.

Review is up on PR 101: https://github.com/abundantbeing/hermes-browser-extension/pull/101#issuecomment-5590506410

Not approved. The CSS fix is the right idea (dock-aware max-height plus shrinking/scrolling rows), and current main still has the 176px bug. GitHub has the PR as not mergeable (dirty, 1 ahead / 35 behind main), and there are no check runs on the head SHA.

Blocker: rebase onto main, then re-check the changelog and sidepanel.css hunks. After that, the usual visual pass in a tall side-panel composer plus the Assist picker.

---

Reviewed signature: `ed4c0b29ba1f8a4252fa2fe30b59d306e7ca7872de152c82eaa6825707df49b5`

abundantbeing:
Thanks @qinxianhahaha! This is on main now as commit 8d686c5. Bottom-anchored popovers reserve the live composer dock height instead of the fixed 176px guess, and the model list and runtime options shrink and scroll so the search box and provider selector stay inside the viewport. Credit is in the commit trailer and the CHANGELOG, plus a regression test covering the floating rules. Closing since the fix has landed; please reopen or file a new issue if anything regresses.

abundantbeing:
## Fix Update

Hermes Browser PR #101 Model Picker Viewport Fix

PR #101 is now on `main` as `8d686c5`.

The model picker now anchors against the live composer dock height instead of a fixed guess. Its model and runtime rows can shrink and scroll, keeping the search box and provider selector inside the viewport. The shared context popover follows the same boundary rule.

abundantbeing:
## Fix Update

Hermes Browser PR #101 Model Picker Viewport Fix

PR #101 is now on `main` as `8d686c5`.

The model picker now anchors against the live composer dock height instead of a fixed guess. Its model and runtime rows can shrink and scroll, keeping the search box and provider selector inside the viewport. The shared context popover follows the same boundary rule.

PR #104, opened 2026-09-14
fix(assist): keep the inline launcher off the text in full-width composers

### LINKED ISSUE #96: Hermes Assist launcher covers text in Messenger composer (inside-end placement over full-width contenteditable)
## Summary

The **Hermes Assist inline launcher icon** is positioned **inside the message composer** on facebook.com / messenger.com and sits on top of the text being typed, so the end of the draft is hidden behind the logo.

Two screenshots show the bug on Messenger (dark theme):

| State | Screenshot |
|---|---|
| Empty composer — launcher rendered inside the input, right after the `Aa` affordance | launcher inside the empty composer |
| Typed text — launcher covers the tail of the typed message and pushes the caret area | launcher covering typed text |

## Extension Version

0.3.0 (also reproduced on current `main` b1426ca)

## Browser & OS

Brave (Chromium) on Windows 11

## Affected Surface

Hermes Assist (inline floating widget on web pages)

## Description of the Problem

On facebook.com / messenger.com the composer is a large single `<div contenteditable>` that fills the whole message bar. Hermes Assist anchors its launcher to the focused editable with the `inside-end` strategy: the 36×36 logo is placed **inside the field**, at `right - launcherSize - inset` / vertically centered. Because the editable is the entire composer bar, the icon overlaps the region where text is typed and rendered — the end of the draft (and the caret) disappear behind the logo. This makes typing unpleasant on Messenger and hides part of what you already wrote.

Same-class risk applies to any site whose focused editable is a full-width bar (many chat apps), since the default `['inside-end']` placement is only safe for multi-line editors with empty right-side space.

## Steps to Reproduce

1. Brave (Chromium) with the extension loaded unpacked; gateway connected (inline assist enabled).
2. Open messenger.com conversation (dark theme), click into the message composer (the `Aa …` bar).
3. Note the Hermes launcher icon inside the composer, near the right end of the input.
4. Type a long sentence. The tail of the text and the caret run under the icon.

## Expected Behavior

The launcher should never occlude text in the focused editable:

- For full-width composer bars, fall back to a placement **outside** the editable (e.g. outside the right edge, or above/below) so the draft stays fully readable, or
- push the icon to a non-overlapping corner/edge (the right side of the composer can be dense — emoji/send buttons — but the launcher should at least not sit on top of the text lane), and/or
- treat `facebook`/`messenger` like `chatgpt` (currently the only adapter that requests `outside-end`/`outside-start`/`above-end`/`below-end`) — i.e. add them to the list that gets detached placement, or make `outside-end` the default for bar-like editables.

## Root-cause hint (from reading the code on `main` b1426ca)

- `extension/content-extractor.js` `resolveInlineSiteProfile()` sets `placement.preferred` to `['outside-end','outside-start','above-end','below-end']` **only** for `adapterId === 'chatgpt'`; every other adapter (including `facebook`) falls through to `['inside-end']`.
- `extension/content-inline-helper.js` `position()` then calls `policy.inlineLauncherPlacement(rect, viewport, { preferred })`, which for `inside-end` computes `inlineLauncherPosition(targetRect …)` → `left: right - launcherSize - inset`, i.e. **inside** the editable's bounding box.
- For Messenger the classified editable is the whole composer bar (full width, ~single line), so "inside the editable, right end" lands exactly on the typing lane. There are no `obstacleElements` for messenger that would block the overlap (`rectsOverlap` is only consulted for the candidate strategies, and `inside-end` candidates are never rejected for overlapping the target's own text area).

Happy to test a fix (e.g. adding `facebook`/`messenger` to the detached-placement adapters, or excluding the launcher rect from the editable's text lane when the field is a short full-width bar).

<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #96</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @khoalx18</code> (replace <code>20</code> with the amount, and <code>@khoalx18</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

abundantbeing:
## What this fixes

On facebook.com and messenger.com the whole message composer is one full-width `<div contenteditable>`, and the Assist launcher landed inside it at `right - launcherSize - inset`, on top of the draft and the caret area (#96).

The root cause is exactly what @khoalx18 found by reading the code: `resolveInlineSiteProfile()` only gave ChatGPT an outside-first preference list, and every other adapter fell through to `['inside-end']`, which by definition sits inside the editable's bounding box.

## The change

Two parts, both in the placement path:

1. **Every adapter now leads with outside placements.** `site-adapters.mjs` gives non-ChatGPT adapters `['outside-end','outside-start','above-end','below-end','inside-end']`. ChatGPT keeps its existing list unchanged.
2. **The policy refuses to cover the field.** `inlineLauncherPlacement()` now runs two passes: outside strategies first, and any candidate that still intersects the field's own text box is rejected. `inside-end` is only reached when no outside placement fits, which preserves the compact-field case. Candidates that only just poke out of the safe area are clamped back in rather than rejected, so an edge-to-edge composer cannot exhaust the outside placements and fall back inside itself.

`content-extractor.js` is regenerated from the lib sources by `scripts/build-content-runtime.mjs`.

## Tests

New `tests/inline-launcher-placement.test.mjs` covers the reported geometry and the behaviours that had to stay intact:

- a full-width composer with room to the right places the launcher outside it, never intersecting the editable box
- an edge-to-edge composer at the bottom of the viewport still avoids the draft
- a compact, edge-to-edge field that leaves no room outside still uses `inside-end`
- callers that allow only `inside-end` keep the original behaviour
- an outside placement that lands on an obstacle falls through to the next strategy

Two existing assertions in `tests/site-adapters.test.mjs` encoded the old default (`inside-end` first for every non-ChatGPT adapter) and were updated to the new contract.

## Verification

```
npm run test        -> # tests 1762 / # pass 1762 / # fail 0   (was 1757)
npx eslint . --quiet -> exit 0
npm run build        -> Self-contained check passed, Built unpacked extension
npm run build:firefox-> Self-contained check passed, Built Firefox extension
```

`npm run test:e2e:loaded` exercises the live extension in Chrome and asserts the old `inside-end` geometry in its launcher-placement block, so that block needs a run with a real browser before this merges.

Fixes #96

Thanks to @khoalx18 for the precise root-cause hint and the screenshots, which made this a short path from report to fix. This repository uses Opire rewards, so comment `/claim #96` here once this lands.


abundantbeing:
Merged. Visual of the change:

Before and after launcher placement

Drawn from the shipped placement policy: before is the old `inside-end`-only default, after is the shipped preference selecting `outside-end` for the same rects.

PR #107, opened 2026-09-15
feat(sidepanel): one-click cards for returned files (Open, Open on computer, Save)

abundantbeing:
## What

When a turn produces a file, the transcript used to answer with nothing but the path it sits at. Every returned file now renders as a card — file name, type badge, and the actions that actually work:

| Action | Kinds | What it does |
|---|---|---|
| **Open** | pdf, html, image, text, csv, json, markdown, video, audio | Fetches the bytes over the extension's authenticated dashboard transport (`/api/files/download` with the session token) and shows them from a blob URL in a new tab, so a PDF or text page renders immediately. |
| **Open on computer** | every kind | `downloads.download(...)` → waits for the download to complete → `downloads.open(id)`, so the OS default app (Excel, Word, Acrobat) launches the file. |
| **Save** | every kind | `downloads.download({ saveAs: true })` — the browser's save dialog. |

Kinds a browser cannot preview (xlsx, docx, pptx, zip, unknown extensions) offer Open on computer and Save only — never a fake Open.

## Where

- Cards appear in the **side panel** and in the **full tab**, for a `MEDIA:` tag, a `` reference, or a path written as ordinary text ("Saved the report to C:\...\report.pdf").
- The markdown renderer now emits a button-less chip for a returned file instead of the misleading "Generated image unavailable in this browser response." line. The chip carries the path; the surfaces upgrade it into the card once the bytes are proven readable.
- The pre-existing honest filename chip for unreadable media paths is unchanged, and images/videos keep the media hydration they already had.

## Honesty rules

- One `HEAD` probe (`probeArtifactFileSource`) against the same authenticated route decides whether a card may offer any button. HTTP 403/404/413, an unreachable dashboard, an empty file, or a rejected session token disable **all three** actions, draw a dashed boundary, and print the exact reason on the card.
- A card stranded unreadable (the dashboard was not discovered yet, the gateway was briefly down) heals itself on the next render once the file can be read.
- A click that fails at runtime reports the real reason on the card (and in the composer status on the full tab) instead of pretending success.
- HTML previews go through a blob URL that inherits the extension page's CSP (`script-src 'self'`, no inline allowance), so a previewed page renders its markup and styling without executing scripts.
- No permission changes: `downloads` was already in the manifest and is still only exercised by an explicit user click.

## Files

- `extension/lib/artifact-actions.mjs` (new) — pure classifier: name/extension/kind/badge/MIME/viewability, per-kind action plan, failure notices, gateway download URL, plain-text path extraction.
- `extension/lib/artifact-card.mjs` (new) — chip markup, card DOM, shared hydration (chips + text paths, de-duplication, per-message budget, self-healing).
- `extension/lib/media-source.mjs` — `resolveArtifactFileSource` and `probeArtifactFileSource` on the existing authenticated dashboard transport.
- `extension/lib/common.mjs` — markdown emits the chip for a returned file.
- `extension/sidepanel.js`, `extension/app.js` — surface wiring (labels, handlers, hydration).
- `extension/sidepanel.css`, `extension/app-parity.css` — card styling on theme tokens (hover keeps a visible boundary, no hardcoded white).
- `tests/artifact-file-cards.test.mjs` (new) — 26 tests: kind/viewable/MIME/action matrix, name and path extraction, byte-fetch and probe failure modes, card DOM, hydration/fallback/healing, surface wiring, CSS.
- `CHANGELOG.md`, `README.md`, `PERMISSIONS.md` — docs; the `downloads` row now describes Open on computer and Save.

## Verification

- `npm run test` → **1788 passing, 0 failing** (`main` baseline is 1762 + 26 new).
- `npx eslint . --quiet` → exit 0.
- `npm run build` → `Built unpacked extension: D:\Documents\HERMES_BROWSER\dist`, self-contained check passed.
- `npm run build:firefox` → `Built Firefox extension: D:\Documents\HERMES_BROWSER\dist\firefox`, self-contained check passed.
- Focused: `node --test tests/artifact-file-cards.test.mjs` → 26 tests, 26 pass, 0 fail.

## Not verified headlessly

- `downloads.open` launching the real OS application (Excel, Word, Acrobat) needs a live browser and OS; the code path is exercised up to `downloads.download` plus the completion wait, and the wait/listener logic is unit tested, but the OS hand-off itself is not asserted.
- Whether each browser renders every viewable kind inline from a blob URL (Chromium's PDF viewer, plain text, CSV) — no live browser run in this environment.
- The exact CSP inheritance of a blob document could not be asserted here; the extension manifest declares `script-src 'self'` with no inline allowance, which is what the comment relies on.

## Notes

- `npm run check:manifest` fails on `main` as well: six unowned `SUBAGENTS` strings in `sidepanel.html`/`app.html` from the existing subagent-stack feature. This PR does not touch those files.
- No images or figure assets are attached to this PR.


abundantbeing:
Visual of the change:

Before: a bare file path in the reply. After: clickable file cards with Open, Open on computer and Save

Both cards in the right panel are built by this PR's own `buildArtifactFileCard` and `artifactActionPlan`, not by hand: the PDF gets `Open`, `Open on computer` and `Save`, and the spreadsheet correctly gets `Open on computer` and `Save` only, because no browser can render an xlsx inline.

### One thing to decide before merging
`artifactFileDownloadUrl` puts the session token in the query string (`/api/files/download?path=...&token=...`). That is deliberate: it is the only way the browser's own download machinery can read the file, since `downloads.download` cannot set headers, and the dashboard route accepts a query token. The fetch path uses the `X-Hermes-Session-Token` header instead. The tradeoff is that the URL, token included, can show up in the download record at `chrome://downloads` and in the download history for that entry. Worth a conscious call. The alternative is fetching to a blob and downloading the object URL, which keeps the token out of any URL but holds the file bytes in memory.

abundantbeing:
## Hermes Agent Review

> **Notice:** This review is automated and AI-generated.

## Summary

PR #107 introduces interactive returned-file cards for the sidepanel and full-tab views, replacing plain text paths and fallback chips with action buttons (`Open`, `Open on computer`, and `Save`). While the classification logic and UI state design are well-structured, the PR cannot be merged in its current state due to critical permission omissions, browser user-gesture runtime failures, security exposure from rendering untrusted HTML inside the extension origin, and a truncated diff.

## Blocking issues

1. **Missing `downloads.open` manifest permission**
   `browserApi.downloads.open()` requires the `"downloads.open"` permission in both Chrome and Firefox manifests in addition to `"downloads"`. The PR updates `PERMISSIONS.md` but does not add `"downloads.open"` to `manifest.json`. Calling `downloads.open` at runtime will fail with a permission error.

2. **User activation loss on asynchronous `downloads.open`**
   Chromium strictly requires an active user gesture (transient user activation) to invoke `chrome.downloads.open()`. In `openArtifactCardOnComputer`, the code awaits `resolveFulltabArtifactContext()`, awaits `downloads.download()`, and then waits for the download to finish via `waitForFulltabArtifactDownload()` (up to 120 seconds). By the time the download completes, the original click gesture has expired, causing `downloads.open()` to throw `User gesture required`.

3. **Untrusted HTML loaded in extension origin via blob URLs**
   `html` is classified under `VIEWABLE_KINDS`, and `openArtifactCardInBrowser` resolves the file bytes into a blob URL created by `URL.createObjectURL` inside the extension page context. The resulting `blob:chrome-extension://<id>/...` URL is opened directly in a new tab. This renders untrusted HTML inside the privileged extension origin, exposing extension `localStorage`, `indexedDB`, and internal messaging contexts. Even with Manifest V3 CSP restrictions on inline scripts, untrusted HTML files produced by agent runs or external workspaces must not be loaded directly under the extension origin. HTML previews should either be treated as non-browser-viewable artifacts or isolated in an opaque origin sandbox.

4. **Arbitrary executable execution via "Open on computer"**
   `artifact-actions.mjs` classifies `.exe`, `.msi`, `.dll`, `.bin`, `.iso`, and `.dmg` as `binary`, and `artifactActionPlan` automatically enables `open-on-computer` for all non-blocked files. Exposing a one-click local OS launch mechanism for generated executable binaries and script formats presents an unnecessary execution risk. "Open on computer" should be restricted to safe document and media formats.

5. **Incomplete submission and diff truncation in `sidepanel.js`**
   The diff in the event payload terminates abruptly in `extension/sidepanel.js` inside `resolveArtifactDashboardContext` at `re`. If this reflects the branch contents, `sidepanel.js` contains a fatal syntax error. Additionally, `tests/artifact-file-cards.test.mjs` is completely absent from the diff.

## Suggestions

- **Explicit blob MIME typing:** `resolveArtifactFileSource` consumes `await response.blob()` directly. If `/api/files/download` serves files with `Content-Type: application/octet-stream` or attachment headers, the browser will force a download instead of displaying PDFs or text inline. Slice or reconstruct the blob using `artifactMimeForExtension(extension)`.
- **Blob revocation lifecycle:** `FULLTAB_ARTIFACT_BLOB_TTL_MS` revokes the blob URL after 120 seconds. If a user refreshes or prints a large document tab after two minutes, the source URL will be gone. Consider revoking on tab closure or extending the timeout.
- **Manifest check errors:** Resolve the pre-existing unowned strings in `sidepanel.html` and `app.html` blocking `npm run check:manifest`.

## Checks to run

- **CI check suite verification:** The PR description reports 1,788 passing tests, but the event payload contains no CI run status or check suite proofs. Merging cannot proceed without verified CI status.
- **Live browser download flow:** Manually test `Open on computer` in a live browser to confirm permission declaration and verify how the browser handles user gesture requirements.
- **Manifest validation:** Run `npm run check:manifest` to verify manifest syntax and permission declarations once `downloads.open` is added.
- **Test suite execution:** Run `node --test tests/artifact-file-cards.test.mjs` and full regression tests once the complete implementation of `sidepanel.js` and the test file are supplied.

---

Reviewed signature: `8d9010728de864b81ed49eecc9c3ac56c8b812c371aee75eee743728e248e8ec`

abundantbeing:
Polish pass pushed in `040b4f6`.

### One component, nine themes
The card in nous light, nous dark, mono light, mono dark, cyberpunk dark and senter space light

Before: a bare path. After: the card

What changed: the card now carries a **file-family tint and a distinct boundary per kind** (documents, spreadsheets, media, archives, unmapped) so the five buckets are tellable apart without leaning on colour, the extension is de-emphasised in the filename by size and weight rather than a faded colour, the meta line rides the ink token at 0.8 because the muted token was too soft on the light palettes, the primary action is the filled action tile with the rest outlined, and a chip hydrates into the card with a short reveal that goes instant under `prefers-reduced-motion`.

### Download route
The download path now prefers a **blob fetched over the authenticated header transport**, and only falls back to the query-token URL when that route fails, so the session token no longer rides a URL by default. The fallback is deliberate and commented with its reason.

Gate on `040b4f6`: 1791/1791 tests, eslint exit 0, both builds green, `check:content-runtime` current.

PR #108, opened 2026-09-16
Close the update reload loop: detect a rebuilt dist/, offer Reload now, and check once a day

abundantbeing:
Three changes to the update card, all for load-unpacked installs. No manifest permission was added or changed (`alarms`, `downloads`, `storage` were already declared).

## 1. Close the reload loop
The panel now captures the build identity it booted from (`version`/`commit`/`shortCommit`/`builtAt`/`dirty` read from `build-info.json` over an extension URL) and re-reads that same file with `cache: 'no-store'` at boot, again shortly after boot, on focus/visibility, on a 60-second interval, and as soon as a turn that started from the update prompt settles. When the identity changes — a new commit, or a rebuild that only moved `builtAt` — the update card shows a persistent **A newer build is on disk (built <time>). Reload to run it.** line with a **Reload now** button that calls `runtime.reload()`. The extension is never reloaded by a timer: the click is the only call site.

## 2. Honest no-checkout path
The update dialog's install note now says the in-place route needs a local checkout of the repository on this machine, and the note area carries a **Download the latest release** link. The prepared agent prompt stops and hands over that same releases URL instead of cloning a checkout when none exists on the machine, and it no longer asks the agent to reload through computer-use — it points at the panel's **Reload now** control. The dirty-checkout stop, the `npm run build` step, and the verify-before-claiming-success rule are unchanged.

## 3. Silent once-a-day check
On boot, if the cached result in extension storage is older than 24h, the panel runs the same version/commit check the **Check** button runs — no dialog, no toast, silent on failure — then writes the status line, caches the result and timestamp, and marks **Update** as the primary action (`data-update-action="primary"`) when this build is behind main. The manual Check button is never rate-limited by this window.

## Where the new rules live
`extension/lib/update-flow.mjs` holds the pure parts (build identity compare, the reload-pending notice, the 24h window rule, the cache entry shape, the releases constant, the primary-action rule, the install-note text, and the prepared prompt builder) so `tests/update-flow.test.mjs` asserts the real strings and arithmetic with no DOM, including an executed slice of the panel's reload watch (baseline capture, change detection, throttle, post-turn recheck, and that only the click calls `runtime.reload`). `tests/full-tab.test.mjs` now reads the prepared prompt from the module instead of the panel source.

## Limits
- A real `runtime.reload()` re-reading a rebuilt `dist/` from disk cannot be asserted headlessly; the tests cover the identity comparison, the offer state, and the single-reload call site.
- The in-place route still needs a local checkout plus a connected Hermes agent. A machine without a checkout gets the releases download link, not an install.
- The automatic check reuses the manual check path, so a dirty build still does the source-blob comparison (one extra GitHub API call, at most once a day) — no second, divergent status line.

## Verification
`node --test tests/update-flow.test.mjs` — 30 tests, 30 pass, 0 fail.
`npm run test` — 1821 passing, 0 failing (baseline 1791 + 30).
`npx eslint . --quiet` — exit 0.
`npm run build`, `npm run build:firefox`, `npm run check:content-runtime` — all clean.


abundantbeing:
Visual of the three new states:

Update available, newer build on disk, and the no-checkout fallback

Everything in the shot is the shipped artifact: the panels are built from the real `sidepanel.html` markup and `sidepanel.css`, and the strings come from `extension/lib/update-flow.mjs` at render time, so the figure cannot drift from the code.

- **Update available.** The silent daily check renders the status line and promotes Update to the filled primary action, so a user who never presses Check still sees that an update exists.
- **Newer build on disk.** The panel re-reads its own `build-info.json` (boot, +5s, focus, every 60s, and the moment the update turn settles) and shows the persistent dashed line. `Reload now` calls `browserApi.runtime.reload()`, which re-reads the unpacked files from disk. `runtime.reload` has exactly one call site in the panel and it is bound to that click, so nothing reloads on a timer.
- **No local checkout.** The dialog states the requirement and carries the release link; the prepared prompt stops with that link instead of cloning the repository, and it no longer depends on computer-use for the reload.

Gate: 1821/1821 tests, eslint exit 0, both builds green, `check:content-runtime` current. The one thing that stays unverified headlessly is the OS-level reality of a reload picking up a rebuilt `dist/`, which needs a live browser.

